PR #12, opened 2018-06-30
Add ability to calibrate multiple static cameras to each other only

Levi-Armstrong:
This is very similar to the extrinsic multi static camera but it only calibrate the cameras to each other where the other also combine the wrist to target calibration. It was found that if you only did the cameras, the position between the cameras change around 1mm. This is expected due to the robot not being globally accurate, so in the case of using a robot this method should be more accurate for locating cameras to each other. Next I will add a new method for wrist calibration where the cameras are all fixed to the first camera.


Levi-Armstrong:
see PR #16

PR #16, opened 2018-07-02
Add  ability for target on wrist and multiple static camera calibration in two steps

Levi-Armstrong:
Due to robots not being globally accurate this PR add two new calibration. The first just calibrations the relationship between cameras and assumes the target is visible by all cameras. The second assumes the relationship between cameras is fixed but can be transformed as a group along with wrist calibration to locate the set of cameras to the robot.

Also, this PR adds a tool that will perform the target on wrist and multiple static camera calibration in two steps to achieve a higher accuracy calibration.

@Jmeyer1292 The class names are not a full description of the calibration being performed so I will create an issue to further discuss this.

### DIFF
--- a/rct_examples/CMakeLists.txt
+++ b/rct_examples/CMakeLists.txt
@@ -64,14 +64,16 @@ add_dependencies(${PROJECT_NAME}_multi_static_camera ${${PROJECT_NAME}_EXPORTED_
 
 target_link_libraries(${PROJECT_NAME}_multi_static_camera ${catkin_LIBRARIES})
 
-# Executable for demonstrating extrinsic cal of multiple static camera only, moving target functionality
-add_executable(${PROJECT_NAME}_multi_static_camera_only src/tools/multi_static_camera_only_extrinsic.cpp)
+# Executable for demonstrating extrinsic cal of multiple static camera multi step, moving target functionality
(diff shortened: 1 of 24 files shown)


Levi-Armstrong:
@Jmeyer1292  I am not sure on the names of the classes. What are your thoughts?

**rct_optimizations::ExtrinsicMultiStaticCameraOnlyProblem**
This calibration solves the problem where a camera is presented to a set of cameras multiple times at various different position in the field of view and solve for the transforms between each camera and the first camera. As in some of the others should moving target be added back?

**rct_optimizations::ExtrinsicMultiStaticCameraMovingTargetWristOnlyProblem**
This calibration solve the problem where the relationship is known between a set of cameras but the relationship between the set of cameras and the robot are not known along with the wrist to target.

Levi-Armstrong:
@Jmeyer1292 I have created a set of utility function as a start to reduce duplicated code in the examples. When you have time please take a look and let me know if you want anything restructured or changed.

Jmeyer1292:
@Levi-Armstrong There are a lot of changes here and I'm not so sure about some of the utilities. Nevertheless, I'm going to merge and make further changes myself.

We'll address the naming issue separately.

PR #28, opened 2019-09-04
Melodic migration

colin-lewis-ut:
Fixed changes in dependency inheritance & issue with install space

### DIFF
--- a/rct_image_tools/include/rct_image_tools/image_observation_finder.h
+++ b/rct_image_tools/include/rct_image_tools/image_observation_finder.h
@@ -2,6 +2,7 @@
 #define RCT_IMAGE_OBSERVATION_FINDER_H
 
 #include "rct_image_tools/modified_circle_grid_target.h"
+#include <vector>
 #include <boost/optional.hpp>
 #include <opencv2/core.hpp>
 #include <rct_optimizations/types.h>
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Three comments:

 - is fba1c6f4 supposed to be in this PR? It seems unrelated.
 - two of the commits were made with email addresses not recognised by GH. You might want to change that.
 - af3761cc appears to have different `GIT_COMMITTER_NAME` and `GIT_AUTHOR_NAME` (and perhaps the email address also)

marip8:
We could probably drop `fba1c6f`. It was a temporary fix; I think the circle min/max areas should probably be parameterized, so I could work on that in a separate PR. Also, it seems like the changes in `ad8669f` already exist in the master branch. Not sure why they show up here as a diff.

@gavanderhoorn the intent here was not to branch this repository for Melodic, just to add a few changes that allow CI build on Melodic

gavanderhoorn:
> the intent here was not to branch this repository for Melodic, just to add a few changes that allow CI build on Melodic

I wasn't worried or accusing you of wanting to branch :)

Just wanted to make sure we were all on the same page.

colin-lewis-ut:
I have removed the circle area commit (to be parameterized in the future) and removed the redundant change to the install space.

PR #29, opened 2019-09-23
Allow customizing circle detector parameters

smith-doug:
This exposes CircleDetector::Params and allows you to optionally specify the values that findObservations will use.  It also allows loading the params from a file.

PR #30, opened 2019-09-23
Reject when findCirclesGrid creates duplicates

smith-doug:
cv::findCirclesGrid will occasionally try to rearrange the detected keypoints and create duplicate points.  I believe this is related to https://github.com/opencv/opencv/issues/4775 .  extractKeyPoints will return successfully but with nonsensical and overlapping points.  This isn't a great solution, but at least it will prevent you from adding a completely broken observation to your data set.

Merge after #29


marip8:
Closing and reopening to trigger CI build

marip8:
@smith-doug thanks!

PR #37, opened 2020-05-13
Optimization testing utilities update

marip8:
This PR updates the optimization testing utilities to improve ease of use for generating simulated correspondences

### DIFF
--- a/rct_optimizations/include/rct_optimizations/types.h
+++ b/rct_optimizations/include/rct_optimizations/types.h
@@ -53,6 +53,13 @@ struct Correspondence2D3D
   {
   }
 
+  Correspondence2D3D(const Eigen::Vector3d& in_target_,
+                     const Eigen::Vector2d& in_image_)
+    : in_target(in_target_)
+    , in_image(in_image_)
+  {
+  }
+
   Eigen::Vector3d in_target;
   Eigen::Vector2d in_image;
 };
@@ -63,6 +70,13 @@ struct Correspondence3D3D
   Correspondence3D3D()
     : in_target(Eigen::Vector3d::Zero())
     , in_image(Eigen::Vector3d::Zero())
+  {
+  }
+
+  Correspondence3D3D(const Eigen::Vector3d& in_target_,
+                     const Eigen::Vector3d& in_image_)
+    : in_target(in_target_)
+    , in_image(in_image_)
   {
   }
 
--- a/rct_optimizations/include/rct_optimizations/types.h
+++ b/rct_optimizations/include/rct_optimizations/types.h
@@ -47,20 +47,30 @@ struct Pose6d
 // Useful typedefs shared by calibrations
 struct Correspondence2D3D
 {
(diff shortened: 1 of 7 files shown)


marip8:
@schornakj can you review when you get the chance?

rct_optimizations/include/rct_optimizations/types.h
```diff
@@ -47,20 +47,30 @@ struct Pose6d
 // Useful typedefs shared by calibrations
 struct Correspondence2D3D
 {
+  Correspondence2D3D()
```
schornakj, line 50: Would it also be useful to have a constructor with fields that initialize `in_target` and `in_image`? For example:
```
Correspondence2D3D(const Eigen::Vector3d& pt_target, const Eigen::Vector2d& pt_image) {...}
```

rct_optimizations/test/extrinsic_multi_static_camera_utest.cpp
```diff
@@ -1,24 +1,12 @@
 #include <gtest/gtest.h>
 #include <rct_optimizations/extrinsic_multi_static_camera.h>
 #include <rct_optimizations/ceres_math_utilities.h>
+#include <rct_optimizations_tests/utilities.h>
+#include <rct_optimizations_tests/observation_creator.h>
 
-void makePoints(std::size_t rows, std::size_t cols, double spacing, std::vector<Eigen::Vector3d>& points)
```
schornakj, line 5: Did this get replaced by the functions in `observation_creator.h`?

rct_optimizations/test/extrinsic_multi_static_camera_utest.cpp
```diff
@@ -46,83 +34,116 @@ void printResults(const rct_optimizations::ExtrinsicMultiStaticCameraMovingTarge
...
+  std::vector<CorrespondenceSet> correspondences;
+};
+
+Observations addObservations(const test::Target &target,
+                             const test::Camera &camera,
+                             const Eigen::Isometry3d &base_to_camera,
+                             const Eigen::Isometry3d &wrist_to_target,
+                             const Eigen::Vector3d &axis)
+{
+  const int n = 20;
```
schornakj, line 49: Could we make `n` configurable?

marip8: Yes, but it doesn't make sense to do it here because it is within the unit test. Ultimately I want to move this function to `observation_creator.h`, where we would add the number of observations as an input parameter. I was hoping to do this after the restructure proposed in #38 and a PR from @colin-lewis-19 coming this week to formalize the generation of poses.

rct_optimizations/test/extrinsic_multi_static_camera_utest.cpp
marip8: Yes and no. This function duplicates an existing `Target` struct in `utilities.h`, so I just replaced it with a `Target` object

PR #38, opened 2020-05-14
Observation Refactor

marip8:
This PR refactors the way that correspondences and observations are structured for easier use in testing utilities and optimization problems.

## Proposed Structure
- Observation[ ] (i.e. observation set)
  - Observation
    - Correspondence[ ] (i.e. correspondence set)
      - Correspondence
        - Point in target
        - Point in image
    - Camera pose (wrist pose if camera is moving, identity otherwise)
    - Target pose (wrist pose if target is moving, identity otherwise)

## Motivation
Currently, a class for the observation does not exist. The optimization problems require vectors of correspondence sets and wrist poses with the assumption they are the same length and are associated by index with one another. It makes more sense to couple these together in a struct since the correspondence set is directly associated with the pose at which the sensor data was acquired.

An additional shortcoming of the current implementation is that the optimization classes do not provide members for the "wrist" poses of the target. A static-camera-moving-target problem is defined a separate problem from moving-camera-static-target, when in fact they can be the same. The proposed change adds a member for the transform to the target. This makes the representation of the observation more complete/unambiguous and allows duplicated cost functions to be consolidated.

The creation of the observation structure will also help improve the creation of simulated calibration data for unit testing of the algorithms

## Example
As an example, I refactored the `extrinsic_3d_camera_on_wrist`, `extrinsic_camera_on_wrist`, and `extrinsic_static_camera` into a single, more unified `extrinsic_hand_eye` problem which supports all those types of calibration and utilizes the new `Observation` structure. I still need to implement a few more unit tests to prove these out, but they should work correctly.

@Levi-Armstrong @schornakj @drchrislewis can you review the proposed [`Observation` class](https://github.com/marip8/robot_cal_tools/blob/e3e4f9b540dba80e113963de5214dfacdc34c9af/rct_optimizations/include/rct_optimizations/types.h) and hand-eye implementation ([header](https://github.com/marip8/robot_cal_tools/blob/e3e4f9b540dba80e113963de5214dfacdc34c9af/rct_optimizations/include/rct_optimizations/extrinsic_hand_eye.h) and [source](https://github.com/marip8/robot_cal_tools/blob/e3e4f9b540dba80e113963de5214dfacdc34c9af/rct_optimizations/src/rct_optimizations/extrinsic_hand_eye.cpp))? I would like to know if this is a change worth adding into the repository, as it would be a breaking change when the optimization problems are updated.

Builds on #37

### DIFF
--- a/rct_optimizations/include/rct_optimizations/types.h
+++ b/rct_optimizations/include/rct_optimizations/types.h
@@ -52,33 +52,33 @@ struct Pose6d
 /**
  * @brief A pair of corresponding features in a N-dimensional sensor "image" and 3D target
  */
-template<Eigen::Index OBS_DIMENSION>
+template<Eigen::Index IMAGE_DIM, Eigen::Index WORLD_DIM>
 struct Correspondence
 {
-  using Set = std::vector<Correspondence<OBS_DIMENSION>>;
+  using Set = std::vector<Correspondence<IMAGE_DIM, WORLD_DIM>>;
   Correspondence()
-    : in_target(Eigen::Vector3d::Zero())
-    , in_image(Eigen::Matrix<double, OBS_DIMENSION, 1>::Zero())
+    : in_image(Eigen::Matrix<double, IMAGE_DIM, 1>::Zero())
+    , in_target(Eigen::Matrix<double, WORLD_DIM, 1>::Zero())
   {
   }
 
-  Correspondence(const Eigen::Vector3d& in_target_,
-                 const Eigen::Matrix<double, OBS_DIMENSION, 1>& in_image_)
+  Correspondence(const Eigen::Matrix<double, IMAGE_DIM, 1>& in_image_,
+                 const Eigen::Matrix<double, WORLD_DIM, 1>& in_target_)
     : in_target(in_target_)
     , in_image(in_image_)
   {
   }
 
-  /** @brief XYZ location of the feature relative to the target origin (meters) */
-  Eigen::Vector3d in_target;
-
   /** @brief N-dimensional location of the feature relative to the sensor */
-  Eigen::Matrix<double, OBS_DIMENSION, 1> in_image;
+  Eigen::Matrix<double, IMAGE_DIM, 1> in_image;
+
+  /** @brief N-dimensional location of the feature relative to the target origin */
+  Eigen::Matrix<double, WORLD_DIM, 1> in_target;
 };
 /** @brief Typedef for correspondence between 2D feature in image coordinates and 3D feature in target coordinates */
-using Correspondence2D3D = Correspondence<2>;
+using Correspondence2D3D = Correspondence<2, 3>;
 /** @brief Typedef for correspondence between 3D feature in sensor coordinates and 3D feature in target coordinates */
-using Correspondence3D3D = Correspondence<3>;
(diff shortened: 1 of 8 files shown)


schornakj:
It might be useful to template observations based on a type that contains more info about the dimensionality of both the world and the image, vs. just switching between 2D and 3D images. I'm picturing being able to have `Observation<Image<2>, World<Cartesian3>>`. Might be overkill for our current use cases, but it would open the door to doing more exotic optimizations later, like if the image data is 2D LIDAR scans.

I think the only thing that changes between the camera-in-hand and target-in-hand situations is the order in which the transforms are multiplied to go from the camera frame to the target, and there's probably a more intuitive way to represent that than by having redundant class member transforms.

marip8:
I just added a couple more commits to implement the unit tests for 2D and 3D cameras, as well as fix for #40

rct_optimizations/include/rct_optimizations/types.h
```diff
@@ -44,46 +49,80 @@ struct Pose6d
...
-struct Correspondence3D3D
+  /** @brief N-dimensional location of the feature relative to the target origin */
+  Eigen::Matrix<double, WORLD_DIM, 1> in_target;
+};
+/** @brief Typedef for correspondence between 2D feature in image coordinates and 3D feature in target coordinates */
+using Correspondence2D3D = Correspondence<2, 3>;
+/** @brief Typedef for correspondence between 3D feature in sensor coordinates and 3D feature in target coordinates */
+using Correspondence3D3D = Correspondence<3, 3>;
+
+// Deprecated typedefs
```
Levi-Armstrong, line 83: I recommend adding the deprecated macros

rct_optimizations/include/rct_optimizations/types.h
```diff
@@ -44,46 +49,80 @@ struct Pose6d
...
   }
 
-  Correspondence3D3D(const Eigen::Vector3d& in_target_,
-                     const Eigen::Vector3d& in_image_)
-    : in_target(in_target_)
-    , in_image(in_image_)
+  Observation(const Eigen::Isometry3d& to_camera_mount_,
+              const Eigen::Isometry3d& to_target_mount_)
+    : to_camera_mount(to_camera_mount_)
+    , to_target_mount(to_target_mount_)
```
schornakj, line 111: Would be preferable to name these to describe both the parent and child frames

marip8: Are you thinking something like `camera_base_to_camera_mount`?

Naming is a little tricky because different optimizations will interpret these transforms differently. For example the hand-eye calibration we make the assumption that the `to_target_mount` and `to_camera_mount_` are relative to the same frame. However in the case where both the target and camera are actuated by kinematic chains we might develop a new calibration algorithm where these transforms are not relative to the same frame and we optimize a transform that connects them (sidebar: maybe we should actually set up the hand-eye calibration like this, and add the ability to choose whether or not that connecting transform is part of the optimization). I initially left the names a little bit vague to represent this

marip8:
I think the additional commits have addressed the concerns brought up in review. I would like to merge this today so we can move forward with other development based off of this. Explicit approval from a reviewer would be appreciated; otherwise I will merge unless there are other comments

PR #39, opened 2020-05-14
Fix eigen build issue

jrgnicho:
Without these changes It fails to find the **Eigen/Dense** header in `rct_examples` and `rct_image_tools` in ROS melodic


marip8:
Should this also be added to `rct_optimizations`?

jrgnicho:
> Should this also be added to `rct_optimizations`?

@marip8 my guess is that the ceres library brings eigen into the build for `rct_optimizations`.  That package built for me as is

Issue #35, opened 2020-03-31
Split into ROS and non-ROS repositories?


schornakj:
I've been using robot_cal_tools with ROS2, and I would like to contribute back some of the work I did [in my weird fork](https://github.com/schornakj/robot_cal_tools/tree/fix/ament-build) to enable this without adversely affecting ROS1 projects that use this repo. Since some of the RCT packages contain ROS nodes, I think we will need to reorganize the structure of this repo.

A reasonable initial approach could be:
- Make `rct_optimizations` and `rct_image_tools` pure CMake packages.
- Move `rct_ros_tools` and `rct_examples` to a new repository with parallel branches for ROS1 and ROS2.

I recall that one of the goals of this project is to integrate it with the [`industrial_calibration`](https://github.com/ros-industrial/industrial_calibration) project, with `robot_cal_tools` providing the underlying optimization libraries and `industrial_calibration` providing ROS nodes. I think restructuring this repo a little bit will actually help with that integration process down the road.

marip8:
We should move forward with the changes to `rct_optimizations` and `rct_image_tools` sooner rather than later. Our current development efforts may make it difficult to reconcile those packages (at least `rct_optimizations`) if we wait too long

schornakj:
@marip8 It looks like we're halfway through this, since `rct_optimization` and `rct_image_tools` are pure-CMake packages now. Seems like a good time for us to strategize.

What do you think would be a good way to handle `rct_ros_tools`? The two important parts of that package are:
- The target loader utilities, which take a ROS1 nodehandle to load target properties from parameters.
- The command line cal app, which is a standalone node that interacts with the parameter server and performs TF lookups.

I was originally thinking that some parts (e.g. the data_set library) could be moved into the non-ROS part of this repo, but on further investigation they ultimately depend on ROS parameter loading so that might not be feasible.

For the time being I'm going to make a separate repo for my current ROS2 port of `rct_ros_tools`, since I'd like to take advantage of the improvements we made this summer to the core calibration libraries without being blocked by this.

marip8:
I think we could move the data and target loading functions into `rct_image_tools` since they primarily deal with loading targets and images. I believe `yaml-cpp` and `xmlrpcpp` are both ROS-independent C++ packages that could be used in a ROS1 or ROS2 context, so we could probably just revise the parameter loader utilities to deal in terms of those objects and string filepaths rather than using ROS1 node handles

PR #42, opened 2020-05-19
Convert rct_optimizations to be a ROS-generic CMake package

schornakj:
The first step at enabling #35. The work I had done previously converted the RCT packages to be ROS2 packages via Ament, which isn't an ideal strategy for us to use here, so this is a fresh take on the task.

- Remove Catkin dependencies, macros, and functions from `rct_optimizations`.
- Change minimum required CMake version from 2.8.3 to 3.5.0.
- Add a set of CMake macros (in `rct_optimizations/cmake/rct_macros.cmake`) to streamline package configuration and test setup. The idea is that the other RCT packages will be able to use these too. These macros are lightly modified from the ones used in [Tesseract](https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract/tesseract_common/cmake/tesseract_macros.cmake).
- Use a more "modern CMake" approach in `rct_optimizations/CMakeLists.txt`. It now provides its libraries through the `rct::rct_optimizations` target.
- Move test source code to a `test` directory, which is added as a subdirectory if the `RCT_ENABLE_RUN_TESTING` flag is `True`.
- Change how `rct_image_tools`, `rct_ros_tools`, and `rct_examples` find and link against `rct_optimizations` to accommodate the fact that it is no longer a ROS package.

TODOs:

- [x] Questions for @marip8: Do the changes I made to the way tests are structured still allow CI to build and run tests? Do we have to add `RCT_ENABLE_RUN_TESTING` to `.github_actions.yml`?

### DIFF
--- a/rct_optimizations/CMakeLists.txt
+++ b/rct_optimizations/CMakeLists.txt
@@ -1,31 +1,10 @@
-cmake_minimum_required(VERSION 2.8.3)
-project(rct_optimizations)
+cmake_minimum_required(VERSION 3.5.0)
+project(rct_optimizations VERSION 0.1.0 LANGUAGES CXX)
 
-add_compile_options(-std=c++11 -Wall -Wextra)
-
-find_package(catkin REQUIRED)
+find_package(rct_common REQUIRED)
 find_package(Ceres REQUIRED)
 
-catkin_package(
-  INCLUDE_DIRS
-    include
-    ${CERES_INCLUDE_DIRS}
-  LIBRARIES
-    ${PROJECT_NAME}
-  DEPENDS
-    CERES
-)
-
-###########
-## Build ##
-###########
-
-include_directories(
-  include
-  ${CERES_INCLUDE_DIRS}
-)
-
-add_library(${PROJECT_NAME}
+add_library(${PROJECT_NAME} SHARED
   # Utilities
   src/${PROJECT_NAME}/eigen_conversions.cpp
   # Optimizations (monocular camera)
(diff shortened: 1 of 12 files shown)


rct_optimizations/CMakeLists.txt
```diff
@@ -42,48 +22,32 @@ add_library(${PROJECT_NAME}
...
-)
+rct_configure_package(${PROJECT_NAME})
 
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-  FILES_MATCHING PATTERN "*.h"
+install(DIRECTORY include/${PROJECT_NAME}
+  DESTINATION include
+  FILES_MATCHING PATTERN "*.h" PATTERN "*.hpp"
   PATTERN ".svn" EXCLUDE
```
marip8, line 48: I would remove these lines. There is currently nothing in this directory we don't want to install and we're not using SVN

rct_optimizations/test/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+#find_package(GTest REQUIRED)
```
marip8, line 1: Remove commented code

rct_optimizations/CMakeLists.txt
```diff
@@ -42,48 +22,32 @@ add_library(${PROJECT_NAME}
...
-  catkin_add_gtest(${PROJECT_NAME}_extrinsic_multi_static_camera_tests test/extrinsic_multi_static_camera_utest.cpp)
-  target_link_libraries(${PROJECT_NAME}_extrinsic_multi_static_camera_tests ${PROJECT_NAME}_test_support)
-
-  catkin_add_gtest(${PROJECT_NAME}_extrinsic_camera_on_wrist_tests test/extrinsic_camera_on_wrist_utest.cpp)
-  target_link_libraries(${PROJECT_NAME}_extrinsic_camera_on_wrist_tests ${PROJECT_NAME}_test_support)
+if(RCT_ENABLE_RUN_TESTING)
+  enable_testing()
+  rct_add_run_tests_target()
+  add_subdirectory(test)
 endif()
```
marip8, line 41: I would also consider separating the running of the unit tests from the building of the tests. I'm finding right now that I want to build the tests often and run individual tests offline. This can be problematic because but they take 2-3 minutes in total to run. I don't want to wait that long every time I build unit tests just to see ones that I don't care about at the time run

schornakj: The macros do support separately building and running tests, I just need to change where they're included in relation to the `RCT_ENABLE_RUN_TESTING` flag.

One challenge is that I have to make sure that GTest can be found and linked against whenever we build the tests, which is a little tricky because its authors have conceptually designed it to need to be downloaded and built from source, vs. providing precompiled binaries.

schornakj:
I've made some changes which I will push up once I get tests running through CI again:
- I added a `rct_common` package, which now contains CMake macros and the GTest external package.
- I changed the CMake arguments for building and running tests. Setting `RCT_BUILD_TESTS=True` will build the tests, and setting `RCT_RUN_TESTS=True` will cause them to be run using `make test` after they're built.

rct_optimizations/CMakeLists.txt
schornakj: This should be addressed in my latest changes.

schornakj:
CI currently fails with the error below, which I think is because the environment used in the Github Action doesn't have `git` installed.

```
Errors     << rct_common:cmake /root/target_ws/logs/rct_common/build.cmake.000.log
  CMake Error at /usr/share/cmake-3.5/Modules/ExternalProject.cmake:1757 (message):
  
error: could not find git for clone of GTest
```

marip8:
That seems odd. Doesn't the Docker use git to clone the repository its building as well as packages from .rosinstall files?

You can specify that additional packages be installed before running the build using the `ADDITIONAL_DEBS` environment variable with `industrial_ci`. Maybe try that with `git`?

schornakj:
You're right, I didn't understand it quite correctly and git is already installed at the top level.

schornakj:
@marip8 I rebased onto the latest master branch and fixed the issue with linking to GTest on the xenial job.

I think that the `rct_optimizations` tests are being, with two caveats:
- Setting `RCT_RUN_TESTS=True` makes the tests run both when building and during the `run_tests` step, which makes the CI process take longer.
- The `run_tests` step doesn't appear to report that the `rct_optimizations` tests ran and succeeded.

PR #43, opened 2020-05-19
Update/pose generation

colin-lewis-19:
Utiility functions for generating camera poses relative to a target


colin-lewis-19:
@marip8 Could you give this a look?

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -0,0 +1,43 @@
...
+{
+/**
+ * @brief genHemispherePose Generates camera poses in a hemisphere pattern
+ * @param target_pose The position/orientation of the target
+ * @param r Radius of the hemisphere
+ * @param theta_cnt The number of points in the theta-wise direction
+ * @param phi_cnt The number of points in the phi-wise direction
+ * @return A vector of camera positions & orientations
+ */
+std::vector<Eigen::Isometry3d> genHemispherePose(Eigen::Isometry3d target_pose, double r, int theta_cnt, int phi_cnt);
```
marip8, line 21: `Eigen::Isometry3d` -> `const Eigen::Isometry3d&` and make your inputs all `const`

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @brief genConicalPose Generates camera positions using a conical template, with the target at the 'point'
+ * @param target_pose Position and orientation of the target
+ * @param r Radius of the cone
+ * @param h Height of the cone (distance to target)
+ * @param theta_cnt Number of observations in a theta direction
+ * @return A vector of camera positions & orientations
+ */
+std::vector<Eigen::Isometry3d> genConicalPose(Eigen::Isometry3d target_pose, int observations, double r, double h);
```
marip8, line 30: Same comment about input parameter types

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @brief genGridPose Generates camera positions in a grid pattern
+ * @param target_pose Position & orientation of the target
+ * @param grid_side number of columns & rows to go into grid
+ * @param spacing Distance betweeon points
+ * @param h Grid distance to target
+ * @return A vector of camera positions & orientations
+ */
+std::vector<Eigen::Isometry3d> genGridPose(Eigen::Isometry3d target_pose, int grid_side, double spacing, double h);
```
marip8, line 40: Same comment about input parameter types

Also, `grid_side` to something a little more descriptive like `grid_size` or `row_count`?

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genHemispherePose(Eigen::Isometry3d target_pose, double r, int theta_cnt, int phi_cnt)
+{
+  // all in the target coordinate system, with z forward
+  std::vector<Eigen::Isometry3d> camera_positions;
+
+  Eigen::VectorXd theta_range = Eigen::VectorXd::LinSpaced(theta_cnt, 0.0, M_PI);
+  Eigen::VectorXd phi_range = Eigen::VectorXd::LinSpaced(phi_cnt, 0.0, M_PI);
+
+  for (long theta_it = 0; theta_it < theta_range.size(); ++theta_it)
```
marip8, line 16: Being picky, but `std::size_t` is a better choice than `long` because it's unsigned

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+  {
+    std::vector<Eigen::Isometry3d> local_scan_path;
+    local_scan_path.reserve(static_cast<std::size_t>(phi_range.size() * theta_range.size()));
```
marip8, line 19: You probably don't need to cast since `size` is already `std::size_t`

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+      // camera may not need to be isometry since we only use the translation portion
+      // x is 'up' in target frame (confirm)
+      Eigen::Isometry3d camera_oriented =
+          rct_optimizations::test::lookAt(camera.translation(), target_pose.translation(), Eigen::Vector3d(1, 0, 0));
+
+      // this vector is still in target spatial coordinates
+      camera_positions.push_back(camera_oriented);
+      // for world frame need target * position
+      // camera_positions.push_back(focus_ * camera_oriented)?
```
marip8, line 39: Remove comments that aren't relevant

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
+#include "rct_optimizations_tests/observation_creator.h"
+#include "rct_optimizations_tests/pose_generator.h"
+#include <stdexcept>
+
+// hemisphere
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genHemispherePose(Eigen::Isometry3d target_pose, double r, int theta_cnt, int phi_cnt)
```
marip8, line 8: Can `target_pose` be a 3D vector?

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genGridPose(Eigen::Isometry3d target_pose, int grid_side, double spacing, double h)
+{
+  // Generates positions in target frame; need to convert to world frame
+  std::vector<Eigen::Isometry3d> camera_positions;
+
+  for (int i = -1 * (grid_side - 1) / 2; i <= (grid_side - 1) / 2; ++i)
+  {
+    for (double j = -1 * (grid_side - 1) / 2; j <= (grid_side - 1) / 2; ++j)
```
marip8, line 81: I would suggest moving this logic out of the loops. It's probably easier and more understandable to make a grid from (0,0) and move it to the center of the grid

schornakj: Another approach that might be clearer would be to generate a list of all the `Vector3d`s representing the different positions within the grid, and then have a separate loop which uses a more generic function to generate each camera position from the grid poses.

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+      // for world frame need target * position
+      // camera_positions.push_back(focus_ * camera_oriented)?
+    }
+  }
+
+  return camera_positions;
+}
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genConicalPose(Eigen::Isometry3d target_pose, int observations, double r, double h)
```
schornakj, line 47: This is also very picky but if we always expect `observations` to be positive then it's better to use an unsigned type.

Also better to make the parameters const references, like Michael says in his other comments.

rct_optimizations/test/src/pose_generator.cpp
colin-lewis-19: I think I have addressed everything but this block. I am not sure I understand the alternatives y'all have presented, and would like to discuss briefly on our call in the morning.

rct_optimizations/test/extrinsic_camera_on_wrist_utest.cpp
```diff
@@ -120,21 +162,42 @@ void run_test(InitialConditions condition)
...
+{
+  run_test(InitialConditions::IDENTITY_TARGET, CameraPattern::CONE);
+}
+
+//TEST(CameraOnWrist, identity_target_start_grid)
+//{
+//  run_test(InitialConditions::IDENTITY_TARGET, CameraPattern::GRID);
+//}
+
+
```
marip8, line 195: Remove commented code

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "rct_optimizations_tests/pose_generator.h"
+#include <stdexcept>
+
+// hemisphere
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genHemispherePose(const Eigen::Vector3d& target_pose, const double r, const unsigned int theta_cnt, const unsigned int phi_cnt)
+{
+  // all in the target coordinate system, with z forward
+  std::vector<Eigen::Isometry3d> camera_positions;
```
marip8, line 11: Pre-allocate (i.e. reserve) memory

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  return camera_positions;
+}
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genConicalPose(const Eigen::Vector3d& target_pose, const unsigned int observations, const double r, const double h)
+{
+  std::vector<Eigen::Isometry3d> camera_positions;
```
marip8, line 51: Pre-allocate (reserve) memory

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return camera_positions;
+}
+
+std::vector<Eigen::Isometry3d>
+rct_optimizations::test::genGridPose(const Eigen::Vector3d& target_pose, const unsigned int grid_side, const double spacing, const double h)
+{
+  // Generates positions in target frame; need to convert to world frame
+  std::vector<Eigen::Isometry3d> camera_positions;
```
marip8, line 77: Pre-allocate (reserve) memory

rct_optimizations/test/src/pose_generator.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  double end_point = ((grid_side -1)/2) * spacing;
```
marip8, line 79: Probably should revise your combined int/double math to in case `grid-1` is odd
```c++
double end_point = ((static_cast<double>(grid_side - 1) / 2.0) * spacing;
```

PR #44, opened 2020-05-19
DH Parameter Kinematic Chain Representation

marip8:
This PR introduces two new items in preparation of developing a kinematic calibration routine:
- DH parameter-based kinematic chain representation
- Observation generation using kinematic chains

There a little effort making the DH chain class compatible with Ceres templates, but it should be straightforward and better addressed in a separate PR

Builds on #38


marip8:
@schornakj any thoughts here?

PR #45, opened 2020-05-22
Make rct_image_tools a pure CMake package

schornakj:
This depends on #42 and continues to address #35.

- Modify `rct_image_tools` to be a pure CMake package, instead of a Catkin package.
- Provide `rct_image_tools` libraries through the `rct::rct_image_tools` target.
- Modify packages that use `rct_image_tools` so that they treat it as a CMake package instead of a Catkin package.
- Link against Eigen using the `Eigen3::Eigen` target.
- Use the macros in `rct_common` for package setup.


marip8:
@schornakj do you know why CI hasn't run for this PR? It looks like your branch has the CI config files. Maybe it needs a rebase or we need to close and reopen this PR?

schornakj:
@marip8 I've read that Github Actions have some poorly-documented limitations that sometimes prevent jobs from running. I need to push up a new commit to trigger the pipeline again.

Levi-Armstrong:
B

schornakj:
@Levi-Armstrong @marip8 The current problem is that the `Eigen3::Eigen` target isn't provided by default on Xenial. I have a section in `rct_image_tools/CMakeLists.txt` that creates this target from the Eigen3 definitions and include directories if it isn't found. However, it looks like packages that link against `rct_image_tools` aren't able to find my `Eigen3::Eigen` target.

What's the right way to solve this, other than duplicating the creation of my `Eigen3::Eigen` target in each package that links against `rct_image_tools`?

schornakj:
@marip8 Added a workaround for the Eigen issues I'd been having in the name of getting some of the outstanding PRs resolved. Should be ready to go.

PR #48, opened 2020-05-28
Remove obsolete extrinsic hand eye optimizations

marip8:
This PR replaces the obsolete hand eye optimizations (extrinsic static camera/moving target, extrinsic 2D camera/static target, extrinsic 3D camera/static target) with the single new hand-eye optimization implementation from #38


marip8:
@schornakj any thoughts about the removal of these old optimizations?

schornakj:
@marip8 The changes you made to the example problems (in particular the fact that you only had to change a handful of type names) demonstrate that this is functionally a drop-in replacement, so I'm OK with officially replacing the two old optimization types with the new one.

PR #49, opened 2020-05-28
Pose Generation Update

marip8:
This PR adds functions for generating observation sets using the pose generators from #43. In order to make the pose generators easier to use with the observation generators, their structure was refactored into and interface with implementations.

The hand-eye unit tests currently only use one of the pose generators. We probably want to devise a way of testing all of the pose generators with the optimizations. I am still trying to figure out a clean way to do this but haven't come across a good solution yet.

Builds on #48

### DIFF
--- a/rct_optimizations/test/include/rct_optimizations_tests/observation_creator.h
+++ b/rct_optimizations/test/include/rct_optimizations_tests/observation_creator.h
@@ -2,6 +2,7 @@
 #define RCT_OPTIMIZATIONS_TESTS_OBS_CREATOR_H
 
 #include <rct_optimizations_tests/utilities.h>
+#include <rct_optimizations_tests/pose_generator.h>
 
 namespace rct_optimizations
 {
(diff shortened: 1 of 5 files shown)


marip8:
@colin-lewis-19 can you review the changes to your pose generators and the observation creation functions?

colin-lewis-19:
This looks good to me

rct_optimizations/test/include/rct_optimizations_tests/observation_creator.h
```diff
@@ -38,14 +39,38 @@ Correspondence3D3D::Set getCorrespondences(const Eigen::Isometry3d &camera_pose,
...
+ * @param true_camera_mount_to_camera - the true transform from the camera mount to the camera
+ * @param camera_base_to_target_base - the transform from the camera base frame to the target base frame (typically identity)
+ * @return
  */
-Eigen::Isometry3d lookAt(const Eigen::Vector3d &origin,
-                         const Eigen::Vector3d &eye,
-                         const Eigen::Vector3d &up) noexcept;
+Observation3D3D::Set createObservations(
+  const Target &target,
+  const PoseGenerator &pose_generator,
```
colin-lewis-19, line 70: At least for the sensor noise qualification, it may be useful to add a stationary pose generator (n poses in the same location) or have the ability to bypass doing the pose generation within the function

schornakj:
This also looks good to me.

rct_optimizations/test/include/rct_optimizations_tests/observation_creator.h
marip8: Is this something you're doing in your PR, or should I add it here?

colin-lewis-19: I worked around it in my PR

PR #50, opened 2020-06-02
Replace instances of deprecated CorrespondenceSet with Correspondence2D3D::Set

schornakj:
I think the deprecated tag on `CorrespondenceSet` was causing issues with CI so I went through and updated uses to the non-deprecated replacement.


marip8:
Should we keep the old typedefs now that you've replaced them?

PR #51, opened 2020-06-03
Camera intrinsic calibration validation

marip8:
This PR adds capability for validating the accuracy of a camera intrinsic calibration.

## Assumptions
Given a set of camera intrinsic parameters and a set of observations:
1. The correspondence set in each observation can be split into two independent correspondence sets
    - s1 = (c[0], ..., c[n/2])
    - s2 = (c[(n/2)+1], c[n])
1. A PnP optimization can accurately* determine the transformation of the origin of each virtual observation set relative to the camera
    - The residual error threshold for the PnP optimization can be specified to account for the noise level of the sensor
1. Although the two virtual correspondence sets contain different features, those features share the same origin. Therefore, the transform from the camera to the origin of both virtual correspondence sets should be the same, given perfect camera intrinsic parameters. For imperfect camera intrinsic parameters, there will be some error between the transforms
1. Calculating the mean and standard deviation between virtual correspondence sets for all observations provides a means of assessing the accuracy of the camera intrinsic calibration

### Using Kinect camera estimated intrinsic parameters
- Positional Error:
  - Mean (m): 0.00836456
  - Std. Dev. (m): 0.00868565
- Angular Error:
  - Mean (rad): 0.0327079
  - Std. Dev. (rad): 0.0242015

This means that the upper bound (mean + 2 sigma) of the transform error from one virtual correspondence set to the other was 25.7 mm and 4.64 degrees (i.e. really bad)

### Using Kinect camera calibrated intrinsic parameters
- Positional Error:
  - Mean (m): 0.00590141
    - Std. Dev. (m): 0.0058142
- Angular Error:
    - Mean (rad): 0.0273891
    - Std. Dev. (rad): 0.0218903

This means that the upper bound (mean + 2 sigma) of the transform error from one virtual correspondence set to the other was 17.5 mm and 4.07 degrees (i.e. still pretty bad)

These results along with @schornakj 's covariance analysis in #46 seem to show that there was probably a lack of data diversity in these images. We should eventually create a data set of really good intrinsic calibration images with which to run this test.

### DIFF
--- a/rct_optimizations/src/rct_optimizations/validation/camera_intrinsic_calibration_validation.cpp
+++ b/rct_optimizations/src/rct_optimizations/validation/camera_intrinsic_calibration_validation.cpp
@@ -2,21 +2,6 @@
 #include <boost/accumulators/accumulators.hpp>
 #include <boost/accumulators/statistics.hpp>
 
-// Specialize std::numeric_limits for the Eigen::Vector3d type
-// Ref: https://stackoverflow.com/questions/7908982/using-boost-accumulators-with-eigenvector-types
-namespace std
-{
-template<>
-struct numeric_limits<Eigen::Vector3d>
-{
-  static const bool is_specialized = true;
-  static Eigen::Vector3d max()
-  {
-    return Eigen::Vector3d::Ones() * std::numeric_limits<double>::max();
-  }
-};
-} // namespace std
-
 namespace rct_optimizations
 {
 Eigen::Isometry3d getInternalTargetTransformation(const Correspondence2D3D::Set &correspondences,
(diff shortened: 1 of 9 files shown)


marip8:
@schornakj this is ready for review when you get the chance

rct_optimizations/src/rct_optimizations/validation/camera_intrinsic_calibration_validation.cpp
```diff
@@ -0,0 +1,120 @@
...
+    problem.camera_to_target_guess = camera_to_target_guess;
+
+    PnPResult result = optimize(problem);
+    if (!result.converged || result.final_cost_per_obs > pnp_sq_error_threshold)
+    {
+      std::stringstream ss;
+      ss << "PnP optimization " << (result.converged ? "converged" : "did not converge")
+         << " with residual error of " << result.final_cost_per_obs << " ("
+         << pnp_sq_error_threshold << " max)";
+      throw std::runtime_error(ss.str());
```
schornakj, line 28: Might be a good place to throw our [`OptimizationException`](https://github.com/Jmeyer1292/robot_cal_tools/blob/bea5fda6ee4a3bdd3f7ea3fcb93a329f88364432/rct_optimizations/include/rct_optimizations/types.h#L128-L132).

PR #52, opened 2020-06-04
Update/sensor noise qual

colin-lewis-19:
Adds a tool for qualifying sensor noise, as well as an implementation of 3d pnp.


rct_optimizations/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@ project(rct_optimizations VERSION 0.1.0 LANGUAGES CXX)
 find_package(rct_common REQUIRED)
 find_package(Ceres REQUIRED)
 
+find_package(catkin REQUIRED COMPONENTS
+  rct_common
+)
```
schornakj, line 9: `rct_optimizations` is a pure CMake package now, so make sure you don't add any ROS-specific stuff back in.

rct_optimizations/include/rct_optimizations/noise_qualifier.h
```diff
@@ -0,0 +1,50 @@
...
+  * @return Noise Statiscics: a vector of standard deviations and the mean pos
+  */
+ NoiseStatistics qualifyNoise2D(const NoiseQualParams2D3D& params);
+
+ /**
+  * @brief qualifyNoise3D This function qualifies 3d sensor noise
+  * @param params 3D image parameters
+  * @return Noise Statiscics: a vector of standard deviations and the mean pos
+  */
+ NoiseStatistics qualifyNoise3D(const NoiseQualParams3D3D& params);
```
marip8, line 48: Please provide a more thorough description of what it means to "qualify" sensor noise and maybe a little bit about the method each function uses, assumptions, etc.. It's important to document how these functions work so it's we remember later down the line

rct_optimizations/include/rct_optimizations/noise_qualifier.h
```diff
@@ -0,0 +1,50 @@
...
+namespace rct_optimizations
+{
+
+/**
+  * @brief The NoiseStatistics struct output structure for image analysis data
+  */
+ struct NoiseStatistics
+ {
+   Eigen::VectorXd mean;
+   Eigen::VectorXd std_dev;
```
marip8, line 14: I'm torn about how to represent mean/std. dev. pairs. I don't like maintaining two vectors that are assumed to be the same length. I also don't really like the approach I took in #51 where I used a `std::vector<std::pair<double, double>>` to represent the pair. Maybe we make a simple new struct with mean and std. dev. variables and then keep vectors of that struct. Any thoughts?

rct_optimizations/src/rct_optimizations/noise_qualifier.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    x_acc(result.camera_to_target.translation()(0));
+    y_acc(result.camera_to_target.translation()(1));
+    z_acc(result.camera_to_target.translation()(2));
+
+    //TODO: Invesitgate using a PoseXd instead?
+    Eigen::Vector3d ea = result.camera_to_target.linear().eulerAngles(0,1,2);
+    r_acc(ea(0));
+    p_acc(ea(1));
+    yw_acc(ea(2));
```
marip8, line 60: Should we check that the optimization converges before adding data?

rct_optimizations/src/rct_optimizations/noise_qualifier.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    x_acc(result.camera_to_target.translation()(0));
+    y_acc(result.camera_to_target.translation()(1));
+    z_acc(result.camera_to_target.translation()(2));
+
+//    similar transforms should give similar eas (caveat: tranforms near singularities are vulnerable)
+    Eigen::Vector3d ea = result.camera_to_target.linear().eulerAngles(0,1,2);
+    r_acc(ea(0));
+    p_acc(ea(1));
+    yw_acc(ea(2));
```
marip8, line 121: Same comment about checking convergence

rct_optimizations/src/rct_optimizations/pnp.cpp
```diff
@@ -44,6 +44,43 @@ struct SolvePnPCostFunc
...
+    T target_pt[3];
+    target_pt[0] = T(in_target_(0));
+    target_pt[1] = T(in_target_(1));
+    target_pt[2] = T(in_target_(2));
+
+    T camera_point[3];  // Point in camera coordinates
+    rct_optimizations::transformPoint(target_angle_axis, target_position, target_pt, camera_point);
+
+    T xyz_image[3];
+    //rct_optimizations::projectPoint(intr_, camera_point, xyz_image); //Does this really need to go?
```
marip8, line 71: Yes, this does really need to go; please delete

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,126 @@
...
+  rct_optimizations::Observation2D3D::Set observations;
+  observations.reserve(obs_cnt);
+
+  //create poses outside the loop; would be better to decalre pose, then look at from there
+  Eigen::Isometry3d target_loc = Eigen::Isometry3d::Identity();
+  target_loc.translate(Eigen::Vector3d(1.0,0.0,0.0));
+  Eigen::Isometry3d p = Eigen::Isometry3d::Identity();
+
+  p.translation() = Eigen::Vector3d(0.05, 0, 0.1);
+  p.linear() << 0, 0, 1, -1, 0, 0, 0, -1, 0;
```
marip8, line 28: Please use the `rotate` function instead. This isn't wrong, but I have a hard time looking at these numbers and understanding how you've changed the pose

marip8, line 28: Does this pose combination actually work? It seems like target is 0.95 meters in front of the camera, but the camera is only ~4" above the plane of the target (presumably with the camera looking straight down)

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,126 @@
...
+    //use camera pose for pose
+    rct_optimizations::Observation2D3D ob;
+    ob.to_target_mount = target_loc;
+    ob.to_camera_mount = p;
+
+    ob.correspondence_set = getCorrespondences(ob.to_camera_mount,
+                                             ob.to_target_mount,
+                                             camera,
+                                             target,
+                                             true);
```
marip8, line 43: If you are going to use `true` for the "all_required" argument, you should consider putting this line inside a `EXPECT_NO_THROW`

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,126 @@
...
+  param.camera_guess = guess;
+
+  rct_optimizations::NoiseStatistics output = rct_optimizations::qualifyNoise2D(param);
+
+  EXPECT_TRUE(output.std_dev(0) < 0.00001);
+  EXPECT_TRUE(output.std_dev(1) < 0.00001);
+  EXPECT_TRUE(output.std_dev(2) < 0.00001);
+  EXPECT_TRUE(output.std_dev(3) < 0.00001);
+  EXPECT_TRUE(output.std_dev(4) < 0.00001);
+  EXPECT_TRUE(output.std_dev(5) < 0.00001);
```
marip8, line 64: If we use a perfect guess and perfect data, should we expect better standard deviations (i.e. 1.0e-15, or std::numeric_limits<double>::epsilon())?

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,126 @@
...
+  param.camera_guess = guess;
+
+  rct_optimizations::NoiseStatistics output = rct_optimizations::qualifyNoise3D(param);
+
+  EXPECT_TRUE(output.std_dev(0) < 0.00001);
+  EXPECT_TRUE(output.std_dev(1) < 0.00001);
+  EXPECT_TRUE(output.std_dev(2) < 0.00001);
+  EXPECT_TRUE(output.std_dev(3) < 0.00001);
+  EXPECT_TRUE(output.std_dev(4) < 0.00001);
+  EXPECT_TRUE(output.std_dev(5) < 0.00001);
```
marip8, line 119: Same comment as above

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+int main(int argc, char **argv)
+{
+  testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
marip8, line 126: This unit test doesn't fully capture everything we want to test. We should be able to add zero-mean Gaussian noise to the observations, and expect the means to be approximately zero and the standard deviations to be on the same order as the standard deviation of the noise

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -0,0 +1,43 @@
...
+      Eigen::Vector3d camera_point = world_point;
+      c.in_target = world_point;
+      c.in_image = camera_point;
+
+      corrs.push_back(c);
+    }
+  }
+
+  setup.correspondences = corrs;
+  setup.camera_to_target_guess = Eigen::Isometry3d::Identity();
```
marip8, line 33: 1. Please use the correspondence generation functions we've already created to make this unit test consistent with the others and to reduce code maintenance
1. Please put the camera in a realistic position for viewing the targets. These tests should be relatively representative of real use-cases (for understandability of the tests, at the very least). I would suggest putting the camera directly over the target by ~1.0 m and rotating it such that the Z-axis of the camera points at the target

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  PnPResult res = rct_optimizations::optimize(setup);
+  EXPECT_TRUE(res.converged);
```
marip8, line 36: We should also expect:
1. The initial error to be extremely low (1.0e-15), since everything about the setup was perfect
1. The final error to be extremely low (1.0e-15), since everything about the setup was already perfect
1. The resulting transform should be almost exactly the same as the input (i.e. `EXPECT_TRUE(res.camera_to_target.isApprox(setup.camera_to_target)`)

rct_optimizations/include/rct_optimizations/noise_qualifier.h
colin-lewis-19: Each Vector currently has 6 elements for x,y,z roll, pitch, yaw. This could be stored in 2 pose6d: one for mean, and one for standard deviation

marip8: Eventually I would like to remove `Pose6d` and use Eigen objects instead, so I wouldn't suggest going that route

colin-lewis-19: Understood. In that case, I think a simple struct comprised of a mean and std dev is probably the best option. Should that be defined in noise_qualifier, or should I put it in types.h for future applications?

marip8:
One more comment: could you move the noise qualification code into a directory inside `include` and `src` called `validation` (instead of `verification`). I want to be consistent with #51

rct_optimizations/include/rct_optimizations/noise_qualifier.h
marip8: For now, just put it inside the noise qualification header. We can amend #51 later to use the same class

rct_optimizations/src/validation/noise_qualifier.cpp
```diff
@@ -0,0 +1,132 @@
...
+      solution_transforms.push_back(result.camera_to_target);
+
+      x_acc(result.camera_to_target.translation()(0));
+      y_acc(result.camera_to_target.translation()(1));
+      z_acc(result.camera_to_target.translation()(2));
+
+      Eigen::Vector3d ea = result.camera_to_target.linear().eulerAngles(0,1,2);
+      r_acc(ea(0));
+      p_acc(ea(1));
+      yw_acc(ea(2));
```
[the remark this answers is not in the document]
colin-lewis-19: I think I am misunderstanding this, but aren't the results of decomposing a rotation matrix into x,y,z components analogous to Euler angles? Or at least susceptible to the same non-uniqueness (ie flipping pi about the x-axis is functionally the same as pi about z + pi about y)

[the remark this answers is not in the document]
marip8: I think you're right about still being subject to non-uniqueness. We just need to get the orientation in a form where you can compare them in your unit tests. Currently you note:
> Euler angles are excluded because of singularities

Angle axis representation (or quaternions for that matter) would be an alternative for getting around this issue

rct_optimizations/test/noise_qualification_utest.cpp
```diff
@@ -0,0 +1,468 @@
...
+  EXPECT_TRUE(output.w.std_dev < 1.5 * stddev);
+
+  //Absolute value of quaternion is taken, since quaternions equal their oppoisite
+  EXPECT_TRUE(abs(output.x.mean - camera_loc.translation()(0)) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.y.mean - camera_loc.translation()(1)) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.z.mean - camera_loc.translation()(2)) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.i.mean) - abs(q_ver.x()) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.j.mean) - abs(q_ver.y()) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.k.mean) - abs(q_ver.z()) < 1.5 * stddev);
+  EXPECT_TRUE(abs(output.w.mean) - abs(q_ver.w()) < 1.5 * stddev);
```
marip8, line 157: I think we can compare position a little bit more accurately. Using the camera projection matrix we can come up with the differentials of (x, y, z) with respect to our perturbed input (u, v)
- delta_x < (z / f_x) * delta_u
  - `dx < (camera_loc.translation.z() / camera.fx()) * stddev`
- delta_y < (z / f_y) * delta_u
  - `dy < (camera_loc.translation.z() / camera.fy()) * stddev`
- delta_z < (f_x * x * (u - c_x)^-2) * delta_u
  - `dz < (camera.fx() * camera_loc.translation.x() * std::pow((u - camera.cx()), -2.0)) * stddev`
  - `u` in this case is obtained by projecting the actual point into the camera

As far as orientation, I don't think we can directly compare it to the pixel noise level (i.e. `stddev`), so I would suggest setting a somewhat arbitrary value as the threshold and comparing distance between the mean and expected quaternion with:

```c++
EXPECT_LT(Eigen::Quaterniond(output.w.mean, output.i.mean, output.j.mean, output.k.mean).angularDistance(q_ver), angular_threshold);
```

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -0,0 +1,81 @@
...
+  setup.correspondences = getCorrespondences(camera_point,
+                                             target_point,
+                                             target);
+
+  setup.camera_to_target_guess = camera_point;
+
+  PnPResult res = rct_optimizations::optimize(setup);
+
+  EXPECT_TRUE(res.converged);
+  //costs are oddly high for 3d, perfect guess
```
marip8, line 70: This would be an indication that your cost function is not correct. If you put in perfect data, your initial cost per observation should be close to the machine epsilon. You should be seeing error on the order of 1.0e-15. Another way to validate is to make a test where the starting condition is not perfect and see if it converges with low residual error

PR #54, opened 2020-06-08
Eigen-based PnP Cost Function and Unit Test

marip8:
This PR adds a unit test for the 2D PnP optimization. I used this as an attempt to understand how to use Eigen objects directly in optimizations for use in the development of future cost functions (i.e. kinematic calibration). Some changes to note:

1. The cost function takes two arguments (camera to target quaternion and position), opposed to the single argument previously (camera to target position and un-normalized axis angle rotation).
1. The cost function uses the `Eigen::Map` class to construct an Eigen object directly from a template data pointer
1. Uses Ceres local parameterization to reduce the 4 quaternion values into the minimal number of variables for optimization (i.e. 3)
    - Supposedly you can create an [SE(3) local parameterization](http://ceres-solver.org/nnls_modeling.html#_CPPv2N5ceres23ProductParameterizationE), such that the cost function input can be a single 6 DOF pose. I wasn't able to figure out how that would work with Eigen objects yet though

Overall writing this cost function using Eigen objects was not difficult, and it makes the functions much more understandable. Also porting this change to the other cost functions would eventually allow us to phase out the custom `Pose6d` to reduce overall code size/maintenance.


marip8:
@schornakj FYI [d5382d4](https://github.com/Jmeyer1292/robot_cal_tools/pull/54/commits/d5382d42cfa87a4a2659b390891c3d4e0b8584bc) was @Levi-Armstrong 's fix for printing CTest output. Hopefully this will help us diagnose unexpected unit test failures more easily

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+using namespace rct_optimizations;
+
+TEST(PNP_2D, PerfectInitialConditions)
+{
+  test::Camera camera = test::makeKinectCamera();
+  test::Target target(5, 7, 0.025);
+
+  Eigen::Isometry3d camera_to_target(Eigen::Isometry3d::Identity());
+  camera_to_target.translate(Eigen::Vector3d(0.0, 0.0, 1.0));
```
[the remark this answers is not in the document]
marip8: > you could have x and y values of the translation be half the target width and height to center the camera over the target.

Good point about the centering the camera.

> Also, I'd like to see an initial offset in the guess for the pose to show that the original values used to create the observations are recovered.

This file contains three unit tests, two of which perturb the initial transform guess

Also interesting points about the covariance. Now that it's merged, I'll give it a shot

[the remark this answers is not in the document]
marip8: The addition of covariance to this one ended up being a little more involved. I'll handle that in a separate PR

rct_optimizations/src/rct_optimizations/pnp.cpp
```diff
@@ -1,37 +1,37 @@
...
 
-    // Transform points into camera coordinates
-    T target_pt[3];
-    target_pt[0] = T(in_target_(0));
-    target_pt[1] = T(in_target_(1));
-    target_pt[2] = T(in_target_(2));
+    Eigen::Map<const Eigen::Quaternion<T>> q(target_q);
+    Eigen::Map<const Vector3> t(target_t);
+    Isometry3 camera_to_target = Isometry3::Identity();
+    camera_to_target = Eigen::Translation<T, 3>(t) * q;
```
schornakj, line 29: Is it possible to map `target_t` directly to an `Eigen::Translation` instead of needing to go through an `Eigen::Matrix<T,3,1>` as an intermediary?

schornakj:
Conceptually this looks good to me! I agree that the Eigen-based approach is more user-friendly than the custom Pose6d type we were using previously.

>     1. The cost function takes two arguments (camera to target quaternion and position), opposed to the single argument previously (camera to target position and un-normalized axis angle rotation).

Strictly speaking the cost function could still take the quaternion and position as a single argument, you'd just need to do some pointer arithmetic when constructing the Eigen mapping. However, I agree with you that it's better to separate them, since it frees the user of the burden of needing to keep track of which indices of a raw pointer correspond to which coefficients in the parameters.

rct_optimizations/src/rct_optimizations/pnp.cpp
marip8: Good thought. I'm not sure if the pointer maps the same way, but I can test it out

marip8: There appears to be a [bug](https://gitlab.com/libeigen/eigen/-/issues/669) with the way that Eigen maps translations from pointers into objects, so it doesn't compile. I think we'll have to leave it as a vector for now. Ultimately I would like to be able to pass the entire matrix to a local  parameterization instead of two separate objects, but I haven't worked that out yet.

PR #58, opened 2020-06-17
Allow PoseGenerators to generate a greater variety of poses

schornakj:
- Add new parameters to `PoseGenerator` implementations for rotation around camera Z-axis and transform offset relative to target origin.
- Allow multiple `PoseGenerator`s to be used to create observations when setting up a given test case.
- Pass around `SharedPtr`s to `PoseGenerator`s (needed to handle collections of `PoseGenerator`s, since the base class is abstract).

### DIFF
--- a/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
+++ b/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
@@ -256,7 +256,7 @@ TYPED_TEST(HandEyeTest, RandomAroundAnswerInitialConditions)
 
 TYPED_TEST(HandEyeTest, RandomAroundAnswerInitialConditions_MoreHemispheres)
 {
-  const std::size_t n = 10;
+  const std::size_t n = 1;
   const std::size_t max_attempts = 2 * n;
   std::size_t count = 0;
 
@@ -265,8 +265,8 @@ TYPED_TEST(HandEyeTest, RandomAroundAnswerInitialConditions_MoreHemispheres)
   offset.translation().x() = 0.1;
 
   std::vector<std::shared_ptr<test::PoseGenerator>> pgs;
-  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, Eigen::Isometry3d::Identity()));
-  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, offset));
+  pgs.push_back(std::make_shared<test::RandomZRotPoseGenerator>(std::make_shared<test::HemispherePoseGenerator>(2.0, 10, 10, 0.0, Eigen::Isometry3d::Identity())));
+  pgs.push_back(std::make_shared<test::RandomZRotPoseGenerator>(std::make_shared<test::HemispherePoseGenerator>(2.0, 10, 10, 0.0, offset)));
 
   while(count < n && count < max_attempts)
   {
--- a/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
+++ b/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
@@ -260,9 +260,13 @@ TYPED_TEST(HandEyeTest, RandomAroundAnswerInitialConditions_MoreHemispheres)
   const std::size_t max_attempts = 2 * n;
   std::size_t count = 0;
 
+
+  Eigen::Isometry3d offset = Eigen::Isometry3d::Identity();
+  offset.translation().x() = 0.1;
+
   std::vector<std::shared_ptr<test::PoseGenerator>> pgs;
-//  pgs.push_back(std::make_shared<test::HemispherePoseGenerator>(2.0, 10, 10));
-  pgs.push_back(std::make_shared<test::HemispherePoseGenerator>(2.0, 10, 10, 0.5 * M_PI));
+  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, Eigen::Isometry3d::Identity()));
+  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, offset));
 
   while(count < n && count < max_attempts)
   {
--- a/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
+++ b/rct_optimizations/test/extrinsic_hand_eye_utest.cpp
@@ -39,7 +39,13 @@ struct ProblemCreator
 
   static ProblemT createProblem(const Eigen::Isometry3d &true_target,
                                 const Eigen::Isometry3d &true_camera,
-                                const test::PoseGenerator& pose_generator,
+                                const std::shared_ptr<test::PoseGenerator>& pose_generator,
+                                const test::Target &target,
+                                const InitialConditions &init);
+
+  static ProblemT createProblem(const Eigen::Isometry3d &true_target,
+                                const Eigen::Isometry3d &true_camera,
+                                const std::vector<std::shared_ptr<test::PoseGenerator>>& pose_generators,
                                 const test::Target &target,
                                 const InitialConditions &init);
 
@@ -52,7 +58,7 @@ template<>
 ExtrinsicHandEyeProblem2D3D ProblemCreator<ExtrinsicHandEyeProblem2D3D>::createProblem(
   const Eigen::Isometry3d &true_target,
   const Eigen::Isometry3d &true_camera,
-  const test::PoseGenerator& pose_generator,
+  const std::vector<std::shared_ptr<test::PoseGenerator>>& pose_generators,
   const test::Target &target,
   const InitialConditions &init)
 {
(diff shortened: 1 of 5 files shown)


schornakj:
The extremely tentative initial result for the hemisphere test case is that adding a second set of observations at the same poses but rotated by `0.5*pi` around the camera-local Z+ axis reduces the correlation between the translation parameters of the camera and wrist transforms, but increases some of the rotation parameters.

Original case (no rotation):

Covariance (camera to wrist):
| | x | y | z | rx | ry | rz |
|-|-|-|-|-|-|-|
| **x** | 0.0669069 | 0.595029 | 0.578816 | 0.277428 | 0.346459 | -0.293642 | 
| **y** | 0.595029 | 0.0536759 | 0.545875 | -0.0104255 | -0.43186 | 0.366025 | 
| **z** | 0.578816 | 0.545875 | 0.0655187 | 0.815311 | -0.227154 | 0.192525 | 
| **rx** | 0.277428 | 0.0104255 | 0.815311 | 0.0924252 | -1.08837e-12 | -2.14972e-12 | 
| **ry** | 0.346459 | -0.43186 | -0.227154 | -1.08837e-12 | 0.0771227 | -0.709471 | 
| **rz** | -0.293642 | 0.366025 | 0.192525 | -2.14972e-12 |  -0.709471 | 0.0282724 | 

New case (more poses rotated by `0.5*pi`):

Covariance (camera to wrist):
| | x | y | z | rx | ry | rz |
|-|-|-|-|-|-|-|
| **x** | 0.021531 | 0.0579856 | 0.145417 | -0.0841606 | 0.391177 | -0.0988148 | 
| **y** | 0.0579856 | 0.027871 | 0.0507486 | -0.701381 | -0.791442 | 0.621118 | 
| **z** | 0.145417 | 0.0507486 | 0.0211778 | 0.595235 | -0.277889 | -0.190204 | 
| **rx** | -0.0841606 | -0.701381 | 0.595235 | 0.045453 | 0.356191 | -0.561755 | 
| **ry** | 0.391177 | -0.791442 | -0.277889 | 0.356191 | 0.0457319 | -0.488505 | 
| **rz** | -0.0988148 | 0.621118 | -0.190204 | -0.561755 | -0.488505 | 0.016959 |

schornakj:
I restructured the `PoseGenerator` struct to allow fully-randomized rotation around camera Z+.

I'll address the implications for covariance in an upcoming PR.

@marip8 Ready for review when you get a chance.

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -25,70 +26,258 @@ struct PoseGenerator
...
+
+  }
+
+public:
+  inline HemispherePoseGenerator(double r_,
+                                 unsigned theta_cnt_,
+                                 unsigned phi_cnt_,
+                                 double z_rot_,
+                                 const Eigen::Isometry3d& target_offset_)
+    : HemispherePoseGenerator(r_, theta_cnt_, phi_cnt_, z_rot_, z_rot_, target_offset_)
```
marip8, line 63: Is this intended to provide a random Z rotation on [`-z_rot`, `z_rot`], or just to lock the Z-rotation at a specific angle?

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -25,70 +26,258 @@ struct PoseGenerator
...
+                                 unsigned long seed_ = std::mt19937::default_seed)
     : r(r_)
     , theta_cnt(theta_cnt_)
     , phi_cnt(phi_cnt_)
+    , target_offset(target_offset_)
+    , mt_rand(seed_)
+    , z_sampler(z_rot_min_, z_rot_max_)
+  {
+
+  }
```
marip8, line 55: It seems a bit backwards to put the implementation specific details of the random Z rotation pose generator in the base class. The generic pose generator classes can be used legitimately without random engines and bounds.

To me it would make more sense if we left `HemispherePoseGenerator` as-is, and added the random engine and bounds to the `RandomZRotHemispherePoseGenerator`. The derived class (random Z rotation) could get the nominal hemisphere poses from the base class, and then apply the random rotations to each pose afterwards

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -25,70 +26,258 @@ struct PoseGenerator
...
-                              double h_ = 2.0,
-                              unsigned n_poses_ = 20)
+protected:
+  inline ConicalPoseGenerator(double r_,
+                              double h_,
+                              unsigned n_poses_,
+                              double z_rot_min_,
+                              double z_rot_max_,
+                              const Eigen::Isometry3d& target_offset_,
+                              unsigned long seed_ = std::mt19937::default_seed)
```
marip8, line 128: Same comments about moving the random elements to the specific random implementation

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -25,70 +26,258 @@ struct PoseGenerator
...
-                           double h_ = 2.0,
-                           unsigned grid_side_ = 10)
+protected:
+  inline GridPoseGenerator(double spacing_,
+                           double h_,
+                           unsigned grid_side_,
+                           double z_rot_min_,
+                           double z_rot_max_,
+                           const Eigen::Isometry3d& target_offset_,
+                           unsigned long seed_ = std::mt19937::default_seed)
```
marip8, line 211: Same comments about moving the random objects to the random implementation

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
```diff
@@ -25,70 +26,258 @@ struct PoseGenerator
    * @param target_origin The position of the target
    * @return A vector of camera positions & orientations
    */
-  virtual std::vector<Eigen::Isometry3d> generate(const Eigen::Vector3d &target_origin) const = 0;
+  virtual std::vector<Eigen::Isometry3d> generate(const Eigen::Isometry3d &target_origin) = 0;
+
+  virtual double getZRot() = 0;
```
marip8, line 31: Is this really an interface-level function? It doesn't seem like this gets called outside the class anywhere. Also it seems like it would be just as simple as getting a random number from the distribution directly instead.

rct_optimizations/test/extrinsic_hand_eye_utest.cpp
```diff
@@ -226,6 +254,66 @@ TYPED_TEST(HandEyeTest, RandomAroundAnswerInitialConditions)
...
+  const std::size_t max_attempts = 2 * n;
+  std::size_t count = 0;
+
+
+  Eigen::Isometry3d offset = Eigen::Isometry3d::Identity();
+  offset.translation().x() = 0.1;
+
+  std::vector<std::shared_ptr<test::PoseGenerator>> pgs;
+  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, Eigen::Isometry3d::Identity()));
+  pgs.push_back(std::make_shared<test::RandomZRotHemispherePoseGenerator>(2.0, 10, 10, offset));
```
marip8, line 269: Should we add more radii (shells) to the hemisphere also?

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
schornakj: I structured it this way in an effort to minimize the amount of duplicate code I was adding. The public constructors of `HemispherePoseGenerator` set ranges for the Z rotation sampler so that it's locked to either zero or a constant value. The constructors for `RandomZRotHemispherePoseGenerator` set the rotation sampler bounds to `0.0` and `2*M_PI` through the protected constructor of `HemispherePoseGenerator`, which enables the random sampling. This way the random sampling pose generators can use the same `generate` function as the base class, and the only thing that is added is the new set of constructors..

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
schornakj: This is intended to lock the Z-rotation to the angle provided in `z_rot_`.

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
schornakj: Agreed, the implementation of this function ends up being very simple anyway. I'll remove this.

rct_optimizations/test/extrinsic_hand_eye_utest.cpp
schornakj: This is one of those tests that gets run 10 times, and it tends to chug if lots of hemispheres are added. I'll change so it just runs once with a larger number of observations.

rct_optimizations/test/include/rct_optimizations_tests/pose_generator.h
marip8: You could probably just use a single class that randomizes the Z-axis orientation for any input "nominal" pose generator

```c++
struct RandomZAxisRotation : PoseGenerator
{
  RandomZAxisRotation(std::shared_ptr<PoseGenerator> generator, double min, double max, double seed)
    : gen(generator)
    , mt_rand(seed)
    , z_sampler(min, max)
  {
  }

  std::vector<Eigen::Isometry3d> generate()
  {
    // Create the nominal poses
    auto poses = gen.generate();

    // Rotate each pose randomly about it's Z-axis
    for(auto& pose : out)
    {
      pose.rotate(Eigen::AngleAxisd(z_sampler(mt_rand), Eigen::Vector3d::UnitZ()));
    }
    return out;
  }

  ...

};
```

schornakj: That's a much more flexible way to do it.

schornakj: Revised to implement randomization the way you describe

PR #60, opened 2020-06-17
Initial DH Chain Kinematic Calibration functionality

marip8:
## Description
This PR implements a number of changes to support a minimum working example of calibration of a kinematic chain based on DH parameters with camera and target

1. Updates the DH chain class to work well with Ceres auto diff cost functions
    - Template specific methods (`getFK` and `getRelativeTransform`) to work easily with auto diff
    - Use Eigen vectors to hold data that needs to be cast between doubles and jets
    - Add the ability to specify DH parameter offsets that will be calibrated by Ceres
    - Added function for getting the DH parameters from a chain in an N x 4 matrix
    - Removed `GaussianNoiseDHTransform` since virtual class functions cannot be templates
1. Defined a kinematic calibration observation structure
1. Created functions for generating kinematic calibration observations
    - Added a function for creating a DH chain with randomly perturbed parameters based on an input chain to replace the `GaussianNoiseDHTransform` class
1. Created initial kinematic calibration algorithm
1. Created unit tests for kinematic calibration algorithm
    - Single kinematic chain with camera transform, target transform, and camera chain to target chain transform

Currently the kinematic calibration converges for the perfect cases represented in the unit test, as well as some perturbed cases that I tried. More development is necessary to finish this algorithm, but that should come in a later PR.

This PR replaces #53 

## TODO
- [x] Drop [e21289d](https://github.com/Jmeyer1292/robot_cal_tools/commit/e21289de77ce0c8008a26a9d5026bd04a2c05395) in favor of the same method added by #54


rct_optimizations/CMakeLists.txt
```diff
@@ -20,10 +20,11 @@ add_library(${PROJECT_NAME} SHARED
   src/${PROJECT_NAME}/camera_intrinsic.cpp
   src/${PROJECT_NAME}/pnp.cpp
   src/${PROJECT_NAME}/multi_camera_pnp.cpp
-  # DH Chain
+  # DH Chain Kinematic Calibration
   src/${PROJECT_NAME}/dh_chain.cpp
+  src/${PROJECT_NAME}/dh_chain_kinematic_calibration.cpp
 )
-target_compile_options(${PROJECT_NAME} PUBLIC -std=c++14)
+target_compile_options(${PROJECT_NAME} PUBLIC -std=c++11)
```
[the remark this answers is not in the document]
marip8: Previously I was using `std::make_unique` which is a C++14 feature; I removed the use of unique pointers in this PR, so I just reverted back to C++11 since 14 wasn't strictly necessary

rct_optimizations/test/dh_chain_kinematic_calibration_utest.cpp
```diff
@@ -0,0 +1,153 @@
...
+  parameters.push_back(q_tm_to_t.coeffs().data());
+  parameters.push_back(t_ccb_to_tcb.data());
+  parameters.push_back(q_ccb_to_tcb.coeffs().data());
+
+  // Create observations
+  KinObservation2D3D::Set observations = test::createKinematicObservations(camera_chain,
+                                                                           target_chain,
+                                                                           camera_mount_to_camera,
+                                                                           target_mount_to_target,
+                                                                           camera_base_to_target_base,
```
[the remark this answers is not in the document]
marip8: Can you clarify your comment? I'm not sure what you mean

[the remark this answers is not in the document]
schornakj: I think there was originally a typo where there was a missing line break between `target_mount_to_target` and `camera_mount_to_base`. I don't see it now so it must have been resolved.

rct_optimizations/include/rct_optimizations/dh_chain_kinematic_calibration.h
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  static std::vector<double *> constructParameters(Eigen::MatrixX4d &camera_chain_dh_offsets,
+                                                   Eigen::MatrixX4d &target_chain_dh_offsets,
+                                                   Eigen::Vector3d &t_cm_to_c,
+                                                   Eigen::Vector3d &aa_cm_to_c,
+                                                   Eigen::Vector3d &t_tm_to_t,
+                                                   Eigen::Vector3d &aa_tm_to_t,
+                                                   Eigen::Vector3d &t_ccb_to_tcb,
+                                                   Eigen::Vector3d &aa_ccb_to_tcb)
```
schornakj, line 88: It would be helpful to either give the arguments more descriptive names or add some `@param` documentation listing what each one corresponds to.

e:  I see that they're documented in more detail in the `optimize` function. I still think it's a good idea to define them here too.

rct_optimizations/src/rct_optimizations/dh_chain.cpp
```diff
@@ -5,98 +5,60 @@ namespace rct_optimizations
...
-  transform.translate(Eigen::Vector3d(r, 0.0, 0.0));
-  transform.rotate(Eigen::AngleAxisd(alpha, Eigen::Vector3d::UnitX()));
-
-  return transform;
-}
-
 double DHTransform::createRandomJointValue() const
 {
   std::mt19937 mt_rand = std::mt19937(std::random_device{}());
   std::uniform_real_distribution<double> dist(min, max);
```
schornakj, line 17: I did a little more reading on `std::mt19937` and the recommended way to set it up is to have just one generator as a class member (`mt_rand` in this case) and use it for all the random number generation that the class instance has to do. This is because it's more computationally-expensive to seed and initialize the generator than to generate a new random number.

However, getting a new random number increments the generator so `createRandomJointValue` couldn't be `const` anymore.

marip8: Is it worth removing the `const` specifier?

marip8: I would prefer keeping the functions `const` because it indicates that calling them won't change the current state of the class which is generally nice to know. I could make `mt_rand` a global variable in that source file, but that doesn't seem like a great practice. If you think `std::mt19937` has issues generating random enough numbers when we are continuously creating new objects, I can change it. If it's just an efficiency thing, then I might be inclined to leave it since this method is only called when generating simulated observations for a chain

schornakj: Technically using a class-member `mt19937` generator to generate a new random number _does_ change the state of the class since the `mt19937` generator gets updated every time it's called. I wish there was some way to indicate the scope of a non-`const` function, since generating random numbers doesn't touch the stored transform info or anything else important.

It's probably not a big deal since in the grand scheme of things we're only generating a few random numbers, but if we ever need to make thousands and thousands then not re-initializing a generator every time may make a difference in the time needed to process. Seems like something we can leave as-is for the time being.

marip8: Maybe using a static `mt19937` in that function is the right way to go. It would allow us to keep the functions const but not have to create a new instance every time the function is called.

schornakj: I think that would be OK.

schornakj:
@marip8 Looks good to me!

PR #63, opened 2020-06-24
PnP Optimization Fix

marip8:
This PR fixes incorrect transformation math from #54 in the projection of target points into the camera frame. It also corrects the `camera_to_target` transform in the unit test, which actually represented the transform from the target to camera


marip8:
@schornakj can you review when you get the chance?

PR #57, opened 2020-06-12
Covariance with local parameterization

marip8:
This PR adds the ability to calculate covariance for parameters that were locally parameterized in the optimization. My test case for development was the Eigen-based PnP optimization in #54. 

# PnP Covariance Results
In these tests the simulated camera was positioned over the center of the target (rather than the origin, which is in the corner) at a 1m standoff 

## 5x7, 0.025m target, perfect initial condition
| | x | y | z | rx? | ry? | rz? |
| --- | --- | --- | --- | --- | --- | --- |
| x |    0.1035 |   0.001181 |     0.5383 | -0.0005392 |          1 |    -0.3187|
| y |  0.001181 |     0.1342 |    -0.5589 |         -1 |  0.0006415 |     -0.207|
| z |    0.5383 |    -0.5589 |   0.007979 |     0.5598 |     0.5394 |   -0.05595|
| rx? |-0.0005392 |         -1 |     0.5598 |      0.067 |  1.457e-14 |     0.2059|
| ry? |         1 |  0.0006415 |     0.5394 |  1.457e-14 |    0.05164 |    -0.3175|
| rz? |   -0.3187 |     -0.207 |   -0.05595 |     0.2059 |    -0.3175 |   0.002711|

## 10x14, 0.025m target, perfect initial condition
| | x | y | z | rx? | ry? | rz? |
| --- | --- | --- | --- | --- | --- | --- |
| x |   0.01265 |   0.004706 |     0.5299 |  -0.002156 |     0.9999 |    -0.3153|
| y |  0.004706 |    0.01627 |    -0.5515 |    -0.9999 |   0.002551 |    -0.2081|
| z |    0.5299 |    -0.5515 |   0.001948 |     0.5553 |     0.5342 |   -0.05263|
| rx? | -0.002156 |    -0.9999 |     0.5553 |   0.008095 | -1.862e-15 |     0.2038|
| ry? |    0.9999 |   0.002551 |     0.5342 | -1.862e-15 |   0.006258 |    -0.3104|
| rz? |   -0.3153 |    -0.2081 |   -0.05263 |     0.2038 |    -0.3104 |  0.0006687|

## 10x14, 0.05m target, perfect initial condition
| | x | y | z | rx? | ry? | rz? |
| --- | --- | --- | --- | --- | --- | --- |
| x |  0.003261 |    0.01837 |     0.5167 |  -0.008358 |     0.9985 |    -0.3293|
[...]

PR #62, opened 2020-06-23
Return CovarianceResult struct instead of raw Eigen matrix

schornakj:
Initial effort at formatting covariance results in an easier-to-use format than Eigen matrices.

- Define a new `NamedParam` type, which describes values that are functions of one or two named parameters.
- Create a `CovarianceResult` struct, which contains vectors of `NamedParam` for the outputs of covariance and correlation calculations. The struct has functions to format itself as a string, and to identify which of its `NamedParam`s have values outside a specified threshold.
- Add new covariance calculation functions which take `string` names as inputs, which allows returning a `CovarianceResult` rather than an Eigen matrix.
- Change the optimizations that were representing covariance matrices as `Eigen::MatrixXd` to instead use `CovarianceResult`.

TODO:

- [x] Fix tests that were disabled by this change
- [x] Add new tests comparing `CovarianceResult` output to Eigen matrix output, to make sure they match

PR #64, opened 2020-06-25
PnP update

marip8:
Per investigation in #57, this PR provides a workaround for using Eigen objects without local paramterization. It also adds a function for calculating the "full" covariance matrix between two parameter blocks (i.e. variance with the parameter block itself and covariance with the other parameter block)

### DIFF
--- a/rct_optimizations/src/rct_optimizations/pnp.cpp
+++ b/rct_optimizations/src/rct_optimizations/pnp.cpp
@@ -1,6 +1,7 @@
 #include "rct_optimizations/pnp.h"
 #include "rct_optimizations/ceres_math_utilities.h"
 #include "rct_optimizations/local_parameterization.h"
+#include "rct_optimizations/covariance_analysis.h"
 
 #include <ceres/ceres.h>
 
@@ -119,6 +120,7 @@ PnPResult optimize(const PnPProblem &params)
   result.initial_cost_per_obs = summary.initial_cost / summary.num_residuals;
   result.final_cost_per_obs = summary.final_cost / summary.num_residuals;
   result.camera_to_target = Eigen::Translation3d(t) * q;
+  result.camera_to_target_covariance = computeFullDV2DVCovariance(problem, t.data(), t.size(), q.coeffs().data(), q_param->LocalSize());
 
   return result;
 }
--- a/rct_optimizations/src/rct_optimizations/pnp.cpp
+++ b/rct_optimizations/src/rct_optimizations/pnp.cpp
@@ -18,16 +18,16 @@ struct SolvePnPCostFunc
   }
 
   template<typename T>
-  bool operator()(const T *const target_q, const T *const target_t, T *const residual) const
+  bool operator()(const T *const target_aa, const T *const target_t, T *const residual) const
   {
     using Isometry3 = Eigen::Transform<T, 3, Eigen::Isometry>;
     using Vector3 = Eigen::Matrix<T, 3, 1>;
     using Vector2 = Eigen::Matrix<T, 2, 1>;
 
-    Eigen::Map<const Eigen::Quaternion<T>> q(target_q);
+    Eigen::Map<const Vector3> aa(target_aa);
     Eigen::Map<const Vector3> t(target_t);
     Isometry3 camera_to_target = Isometry3::Identity();
-    camera_to_target = Eigen::Translation<T, 3>(t) * q;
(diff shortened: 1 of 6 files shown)


marip8:
@schornakj can you review? This is ready to merge but has a few changes that might overlap with #62

rct_optimizations/src/rct_optimizations/pnp.cpp
```diff
@@ -53,16 +53,15 @@ struct SolvePnPCostFunc3D
...
   template<typename T>
-  bool operator()(const T *const target_q, const T *const target_t, T *const residual) const
+  bool operator()(const T *const target_aa, const T *const target_t, T *const residual) const
   {
-
     using Isometry3 = Eigen::Transform<T, 3, Eigen::Isometry>;
     using Vector3 = Eigen::Matrix<T, 3, 1>;
 
-    Eigen::Map<const Eigen::Quaternion<T>> q(target_q);
+    Eigen::Map<const Vector3> aa(target_aa);
```
schornakj, line 61: Nitpicky style preference thing, but I'd prefer if `aa` and `t` were named more verbosely so they can be understood at a glance without needing to follow them through the code.

rct_optimizations/src/rct_optimizations/pnp.cpp
```diff
@@ -85,31 +84,24 @@ namespace rct_optimizations
 PnPResult optimize(const PnPProblem &params)
 {
   // Create the optimization variables from the input guess
-  Eigen::Quaterniond q(params.camera_to_target_guess.rotation());
+  Eigen::AngleAxisd rot(params.camera_to_target_guess.rotation());
+  Eigen::Vector3d aa = rot.angle() * rot.axis();
```
schornakj, line 88: Same comment as above.

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -126,22 +142,14 @@ TEST(PNP_3D, PerfectInitialConditions)
   EXPECT_TRUE(result.camera_to_target.isApprox(target_to_camera.inverse()));
   EXPECT_LT(result.initial_cost_per_obs, 1.0e-15);
   EXPECT_LT(result.final_cost_per_obs, 1.0e-15);
+  checkCorrelation(result.camera_to_target_covariance);
+
+//  Eigen::IOFormat fmt(4, 0, " | ", "\n", "|", "|");
+//  std::cout << "Covariance:\n" << result.camera_to_target_covariance.format(fmt) << std::endl;
```
schornakj, line 148: Do you want to leave this commented out? Might be worthwhile to print during the tests anyway.

schornakj:
> @schornakj can you review? This is ready to merge but has a few changes that might overlap with #62

I'll rebase onto your changes after we merge. #62 adds a function that's similar (but uses different underlying Ceres functions) to the "full" covariance function you added here so you might be interested in taking a look and seeing which approach you prefer.

rct_optimizations/test/pnp_utest.cpp
```diff
@@ -159,7 +159,7 @@ TEST_F(PnP3DTest, PerturbedInitialCondition)
   PnPResult result = optimize(problem);
   EXPECT_TRUE(result.converged);
   EXPECT_TRUE(result.camera_to_target.isApprox(target_to_camera.inverse(), 1.0e-6));
-  EXPECT_LT(result.final_cost_per_obs, 1.0e-12);
+  EXPECT_LT(result.final_cost_per_obs, 1.0e-10);
```
schornakj, line 162: I think we need a better way to evaluate these variable costs in our tests. Maybe we should compare costs across multiple optimizations and check that the mean and variance of all the costs are within some specified range.

marip8:
@schornakj I addressed your comments in the latest commits. I ended up running the randomly initialized optimizations 30 times and setting expectations that the mean error + 3 standard deviations should be less than a (still somewhat arbitrary) threshold. I think this is at least a step in the right direction for setting reasonable expectations for unit tests with randomness

schornakj:
> @schornakj I addressed your comments in the latest commits. I ended up running the randomly initialized optimizations 30 times and setting expectations that the mean error + 3 standard deviations should be less than a (still somewhat arbitrary) threshold. I think this is at least a step in the right direction for setting reasonable expectations for unit tests with randomness

That's reasonable. Looks good to me!

PR #65, opened 2020-06-29
Noise qualification update

marip8:
Changes suggested by @drchrislewis:

- Update to use SVD when calculating eigenvectors for quaternion average as it is more numerically stable 
- Fix the variance calculation to divide by N-1, rather than N

### DIFF
--- a/rct_optimizations/src/rct_optimizations/validation/noise_qualification.cpp
+++ b/rct_optimizations/src/rct_optimizations/validation/noise_qualification.cpp
@@ -1,4 +1,4 @@
-#include <Eigen/Eigenvalues>
+#include <Eigen/SVD>
 #include <rct_optimizations/validation/noise_qualification.h>
 #include <rct_optimizations/pnp.h>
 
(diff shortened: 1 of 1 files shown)


rct_optimizations/src/rct_optimizations/validation/noise_qualification.cpp
```diff
@@ -20,8 +20,8 @@ Eigen::Quaterniond computeQuaternionMean(const std::vector<Eigen::Quaterniond>&
   * M = sum(w_i * q_i * q_i^T)    Eq. 12
   * q_bar = argmax(q^T * M * q)   Eq. 13
   *
-  * The solution of this maximization problem is well known. The average quaternion is
-  * the eigenvector of M corresponding to the maximum eigenvalue.
+  * "The solution of this maximization problem is well known. The average quaternion is
+  * the eigenvector of M corresponding to the maximum eigenvalue."
```
schornakj, line 24: Are these "editorial" kinds of quotes?

marip8: That was a quote that I pulled directly from the paper. I probably should do a better job citing it, but I at least wanted to indicate that that particular comment was not my own

PR #56, opened 2020-06-10
WIP: Update/proclivity

colin-lewis-19:
A function for detecting if target points have been misordered. A pixel threshold can be provided, or outliers can be detected internally by comparing maximum values to interquartile range. Built atop #52

PR #66, opened 2020-06-30
Homography Validation Update

marip8:
This PR revises #56. @colin-lewis-19 can you review?


rct_optimizations/include/rct_optimizations/validation/homography_validation.h
```diff
@@ -0,0 +1,76 @@
...
+  ModifiedCircleGridCorrespondenceSampler(const std::size_t rows_, const std::size_t cols_)
+    : rows(rows_)
+    , cols(cols_)
+  {
+  }
+
+  virtual std::vector<std::size_t> getSampleCorrespondenceIndices() const final override
+  {
+    // Make sure there are at least two times as many points as the number of sample points
+    if ((rows * cols / 2) < 8)
```
colin-lewis-19, line 29: Do we still want to have a variable number of points used, or are we comfortable fixing it to 8?

marip8: I just pushed a change to use the minimum number of points (4, instead of 8). This struct is easy enough to override, so I don't think we necessarily need to make it a parameterizable number of points

PR #69, opened 2020-07-20
Maximum Likelihood

marip8:
This PR adds a template cost function for computing the likelihood of a parameter given its expected mean and standard deviation. The cost function is based on [this example](https://ceres-solver.googlesource.com/ceres-solver/+/master/examples/robot_pose_mle.cc) from Ceres

### DIFF
--- a/rct_optimizations/include/rct_optimizations/maximum_likelihood.h
+++ b/rct_optimizations/include/rct_optimizations/maximum_likelihood.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <Eigen/Core>
+#include <rct_optimizations/types.h>
+
+namespace rct_optimizations
+{
+/**
+ * @brief Cost function for a parameter block based on the expected means and standard deviations for the values in the block.
+ * The inputs are an array of mean values and an array of standard deviations corresponding to the optimization variables in the parameter block
+ *
+ * Ceres provides an example of a maximum likelihood constraint here:
+ * https://ceres-solver.googlesource.com/ceres-solver/+/master/examples/robot_pose_mle.cc#165
+ * blob: 83c0903d11f2558a9eb48884addc44853ffb903b
+ */
+class MaximumLikelihood
+{
+  public:
+  MaximumLikelihood(const Eigen::ArrayXXd &mean_, const Eigen::ArrayXXd &stdev_)
+    : mean(mean_)
+    , stdev(stdev_)
+  {
+    if (mean.size() != stdev.size())
+      throw OptimizationException("Mean and standard deviation matrix are not the same size");
+  }
+
+  template<typename T>
+  bool operator()(T const *const *parameter, T *residual) const
+  {
+    using ArrayXX = Eigen::Array<T, Eigen::Dynamic, Eigen::Dynamic>;
+    Eigen::Map<const ArrayXX> param_array(parameter[0], mean.rows(), mean.cols());
+
+    Eigen::Map<ArrayXX> residual_array(residual, mean.rows(), mean.cols());
+    residual_array = (param_array - mean.cast<T>()) / stdev.cast<T>();
+
+    return true;
+  }
+
+  Eigen::ArrayXXd mean;
+  Eigen::ArrayXXd stdev;
+};
+
+} // namespace rct_optimizations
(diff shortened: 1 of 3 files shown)


rct_optimizations/include/rct_optimizations/maximum_likelihood.h
```diff
@@ -0,0 +1,46 @@
+#ifndef RCT_OPTIMIZATIONS_MAXIMUM_LIKELIHOOD_H
+#define RCT_OPTIMIZATIONS_MAXIMUM_LIKELIHOOD_H
```
schornakj, line 2: Should we be doing `#pragma once` instead?

rct_optimizations/include/rct_optimizations/maximum_likelihood.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+class MaximumLikelihood
+{
+  public:
+  MaximumLikelihood(const Eigen::ArrayXXd &mean_, const Eigen::ArrayXXd &stdev_)
+    : mean(mean_)
+    , stdev(stdev_)
+  {
+    if (mean.size() != stdev.size())
+      throw OptimizationException("Mean and standard deviation matrix are not the same size");
```
schornakj, line 25: Should add an `EXPECT_THROW` test that exercises this.

rct_optimizations/include/rct_optimizations/maximum_likelihood.h
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  template<typename T>
+  bool operator()(T const *const *parameter, T *residual) const
+  {
+    using ArrayXX = Eigen::Array<T, Eigen::Dynamic, Eigen::Dynamic>;
+    Eigen::Map<const ArrayXX> param_array(parameter[0], mean.rows(), mean.cols());
+
+    Eigen::Map<ArrayXX> residual_array(residual, mean.rows(), mean.cols());
+    residual_array = (param_array - mean.cast<T>()) / stdev.cast<T>();
```
drchrislewis, line 34: @marip8 I believe this is missing a sqrt(2) in the denominator. Also from a nomenclature stand point, I don't think this is strictly a maximum likelihood estimate. It certainly adds a cost to straying from the prior. It is however related to the log of the likelihood. It would have made more sense here to use: 1/(sigma*sqrt(2pi)) * e^((-param[i]-mean[i])^2/(2*sigma^2)).
By taking the log, one simplifies the math. The minimum of the log function and the minimum of the likelihood function occur at the same point. However, the cost value is scaled. If you care.

marip8: This probably deserves a little more attention in the future regarding the naming because it is not maximum likelihood estimation, but certainly attempts to maximize the likelihood of the parameter w.r.t the input mean and standard deviation.

I got the math for this from the linked example in the Ceres repository. I was surprised that it was done this way, but it seems to work better than using the Gaussian distribution and log likelihood equations. When I used those equations in the cost function (modified slightly since Ceres does residual minimization rather than maximization), the optimization in the unit test still converged but it wouldn't drive the variables to the mean very well

drchrislewis: @marip8 You can't use the likelihood itself as a cost because you want to maximize likelihood, not minimize a cost. You could create a cost that is the difference in likelihood at the prior, and the current estimate's likelihood. The only thing questionable about what's here is that its units are not likelihood, but its named maximum likelihood. Still if this is the only cost, then it results in a maximum likelihood given the prior meaning it returns the prior(parameter.x = x_bar), the most likely value given your prior knowledge about the variable.

marip8: > You can't use the likelihood itself as a cost because you want to maximize likelihood, not minimize a cost. You could create a cost that is the difference in likelihood at the prior, and the current estimate's likelihood.

Sure, that's what I meant by `(modified slightly since Ceres does residual minimization rather than maximization)`. I implemented this difference strategy with both likelihood and log likelihood, and Ceres was not able to drive the values sufficiently close to the mean

> The only thing questionable about what's here is that its units are not likelihood, but its named maximum likelihood.

I'm open to naming suggestions, but haven't come up with anything that really conveys the purpose of this cost. Maybe `LikelihoodCost`?

PR #73, opened 2020-07-27
Adding ChAruco board detection functionality

jaberkebile:
Using the current Aruco implementation as a reference for ChAruco targets

### DIFF
--- a/rct_examples/src/examples/camera_on_wrist.cpp
+++ b/rct_examples/src/examples/camera_on_wrist.cpp
@@ -113,6 +113,7 @@ int extrinsicWristCameraCalibration()
     obs.correspondence_set.reserve(maybe_obs->size());
 
     // And loop over each detected dot:
+    const std::vector<Eigen::Vector3d> target_points = target.createPoints();
     for (std::size_t j = 0; j < maybe_obs->size(); ++j)
     {
       // The 'target.points' data structure and the observation vector returned by
@@ -120,7 +121,7 @@ int extrinsicWristCameraCalibration()
       // respresents the same dot!
       rct_optimizations::Correspondence2D3D pair;
       pair.in_image = maybe_obs->at(j); // The obs finder and target define their points in the same order!
-      pair.in_target = target.points[j];
+      pair.in_target = target_points[j];
       obs.correspondence_set.push_back(pair);
     }
 
--- a/rct_examples/src/tools/camera_intrinsic_calibration_validation.cpp
+++ b/rct_examples/src/tools/camera_intrinsic_calibration_validation.cpp
@@ -1,5 +1,6 @@
 #include <rct_optimizations/validation/camera_intrinsic_calibration_validation.h>
 #include <rct_ros_tools/parameter_loaders.h>
+#include <rct_ros_tools/target_loaders.h>
 #include <rct_ros_tools/print_utils.h>
 #include <rct_ros_tools/data_set.h>
 
@@ -12,7 +13,11 @@
 #include <opencv2/imgproc.hpp>
 #include <ros/ros.h>
 
-void analyzeResults(const rct_optimizations::IntrinsicCalibrationAccuracyResult &result,
+using namespace rct_optimizations;
+using namespace rct_image_tools;
+using namespace rct_ros_tools;
+
+void analyzeResults(const IntrinsicCalibrationAccuracyResult &result,
                     const double pos_tol, const double ang_tol)
 {
   ROS_INFO_STREAM("Positional Error:\nMean (m): " << result.pos_error.first << "\nStd. Dev. (m): " << result.pos_error.second);
@@ -48,11 +53,11 @@ int main(int argc, char **argv)
 
   try
   {
-    rct_image_tools::ModifiedCircleGridTarget target = rct_ros_tools::loadTarget(pnh, "target_definition");
-    rct_optimizations::CameraIntrinsics intr = rct_ros_tools::loadIntrinsics(pnh, "intrinsics");
+    auto target = TargetLoader<ModifiedCircleGridTarget>::load(pnh, "target_definition");
+    CameraIntrinsics intr = loadIntrinsics(pnh, "intrinsics");
 
     // Attempt to load the data set via the data record yaml file:
-    boost::optional<rct_ros_tools::ExtrinsicDataSet> maybe_data_set = rct_ros_tools::parseFromFile(
+    boost::optional<ExtrinsicDataSet> maybe_data_set = parseFromFile(
       data_path);
     if (!maybe_data_set)
     {
@@ -61,19 +66,19 @@ int main(int argc, char **argv)
     }
 
     // We know it exists, so define a helpful alias
-    const rct_ros_tools::ExtrinsicDataSet& data_set = *maybe_data_set;
+    const ExtrinsicDataSet& data_set = *maybe_data_set;
 
     // Our 'base to camera guess': A camera off to the side, looking at a point centered in front of the robot
-    Eigen::Isometry3d target_mount_to_target = rct_ros_tools::loadPose(pnh, "base_to_target_guess");
-    Eigen::Isometry3d camera_mount_to_camera = rct_ros_tools::loadPose(pnh, "wrist_to_camera_guess");
+    Eigen::Isometry3d target_mount_to_target = loadPose(pnh, "base_to_target_guess");
+    Eigen::Isometry3d camera_mount_to_camera = loadPose(pnh, "wrist_to_camera_guess");
 
     // Lets create a class that will search for the target in our raw images.
-    rct_image_tools::ModifiedCircleGridObservationFinder obs_finder(target);
+    ModifiedCircleGridObservationFinder obs_finder(target);
 
     // Finally, we need to process our images into correspondence sets: for each dot in the
     // target this will be where that dot is in the target and where it was seen in the image.
     // Repeat for each image. We also tell where the wrist was when the image was taken.
-    rct_optimizations::Observation2D3D::Set observations;
+    Observation2D3D::Set observations;
     observations.reserve(data_set.images.size());
     for (std::size_t i = 0; i < data_set.images.size(); ++i)
     {
@@ -93,7 +98,7 @@ int main(int argc, char **argv)
         cv::waitKey();
       }
 
-      rct_optimizations::Observation2D3D obs;
+      Observation2D3D obs;
       obs.correspondence_set.reserve(maybe_obs->size());
 
       // So for each image we need to:
@@ -102,20 +107,15 @@ int main(int argc, char **argv)
       obs.to_target_mount = Eigen::Isometry3d::Identity();
 
       //// And finally add that to the problem
-      obs.correspondence_set = rct_image_tools::getCorrespondenceSet(*maybe_obs, target.points);
+      obs.correspondence_set = getCorrespondenceSet(*maybe_obs, target.createPoints());
 
       observations.push_back(obs);
     }
 
     // Measure the intrinsic calibration accuracy
     // The assumption here is that all PnP optimizations should have a residual error less than 1.0 pixels
-    rct_optimizations::IntrinsicCalibrationAccuracyResult result
-      = rct_optimizations::measureIntrinsicCalibrationAccuracy(observations,
-                                                               intr,
-                                                               camera_mount_to_camera,
-                                                               target_mount_to_target,
-                                                               Eigen::Isometry3d::Identity(),
-                                                               1.0);
+    IntrinsicCalibrationAccuracyResult result = measureIntrinsicCalibrationAccuracy(
+        observations, intr, camera_mount_to_camera, target_mount_to_target, Eigen::Isometry3d::Identity(), 1.0);
 
     // Analyze the results
     // These error tolerances allow the virtual correspondence sets for each observation to deviate from the expectation by up to 1 mm and 0.05 degrees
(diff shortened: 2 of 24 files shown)


rct_image_tools/include/rct_image_tools/charuco.hpp
```diff
@@ -0,0 +1,354 @@
...
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
+
+#ifndef __OPENCV_CHARUCO_HPP__
+#define __OPENCV_CHARUCO_HPP__
+
+#include <opencv2/core.hpp>
+#include <vector>
+#include <opencv2/aruco.hpp>
```
marip8, line 44: This looks like a file that might already be in OpenCV. If it is, let's just include it directly from OpenCV rather than copy-pasting it here

rct_image_tools/include/rct_image_tools/charuco_finder.h
```diff
@@ -0,0 +1,66 @@
+/**
+  * ChAruco gridboard detector, following the same pattern as ModifiedCircleGridObservationFinder.
+  * Author: John Berkebile
+  */
+
+#ifndef CHARUCO_FINDER_H
+#define CHARUCO_FINDER_H
```
marip8, line 7: Please replace your header guards with `#pragma once` or make them more specific (i.e. `RCT_IMAGE_TOOLS_CHARUCO_FINDER_H`)

rct_image_tools/include/rct_image_tools/charuco_finder.h
```diff
@@ -0,0 +1,66 @@
...
+  CharucoGridBoardObservationFinder(const cv::Ptr<cv::aruco::CharucoBoard>& board);
+
+  /**
+   * @brief findObservations Detect marker corner coordinates in the provided image.
+   * @param image Input image, ideally containing an ArUco gridboard.
+   * @return Map matching marker ID numbers to a vector of marker corner coordinates. The vector will contain
+   * four corners defined in the same order as in the output of the function cv::aruco::DetectMarkers()
+   * (e.g. clockwise from the "origin" corner).
+   */
+  boost::optional<std::map<int, std::vector<Eigen::Vector2d>>> findObservations(const cv::Mat& image) const;
```
marip8, line 38: I would remove the `boost::optional` from the return type and just return a map. If there are failures within this function, please throw an exception instead

rct_image_tools/include/rct_image_tools/charuco_finder.h
```diff
@@ -0,0 +1,66 @@
...
+{
+public:
+  CharucoGridBoardObservationFinder(const cv::Ptr<cv::aruco::CharucoBoard>& board);
+
+  /**
+   * @brief findObservations Detect marker corner coordinates in the provided image.
+   * @param image Input image, ideally containing an ArUco gridboard.
+   * @return Map matching marker ID numbers to a vector of marker corner coordinates. The vector will contain
+   * four corners defined in the same order as in the output of the function cv::aruco::DetectMarkers()
+   * (e.g. clockwise from the "origin" corner).
```
marip8, line 36: We actually want to return the ChArUco-interpolated corner points (which are actually the saddle points of the chess board). Otherwise, there is essentially no difference between this and the existing ArUco finder. Please update the documentation and function accordingly

rct_image_tools/src/charuco_test.cpp
```diff
@@ -0,0 +1,60 @@
+#include "rct_image_tools/charuco_finder.h"
+
+int main(int argc, char** argv)
```
marip8, line 3: Please convert this into a unit test instead of an executable per our discussion earlier

rct_image_tools/src/charuco_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+		cv::Mat dots = charuco_finder.drawObservations(m, *maybe_charuco);
+		
+		cv::imshow("out", dots);
+		cv::moveWindow("out", 400, 400);
+		cv:waitKey();
+	}
+	else
+	{
+		std::cout << "Failed to find charuco target\n";
+	}
```
marip8, line 58: No need to visualize targets in a unit test

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+any express or implied warranties, including, but not limited to, the implied
+warranties of merchantability and fitness for a particular purpose are
+disclaimed. In no event shall copyright holders or contributors be liable for
+any direct, indirect, incidental, special, exemplary, or consequential damages
+(including, but not limited to, procurement of substitute goods or services;
+loss of use, data, or profits; or business interruption) however caused
+and on any theory of liability, whether in contract, strict liability,
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
```
marip8, line 37: Please remove the OpenCV license; this package (and contributions to it) are licensed under the Apache 2.0 license

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+	: board_(board)
+{
+
+}
+
+boost::optional<std::map<int, std::vector<Eigen::Vector2d>>>
+rct_image_tools::CharucoGridBoardObservationFinder::findObservations(const cv::Mat &image) const
+{
+	std::map<int, std::vector<Eigen::Vector2d>> map_ids_to_obs_corners;
+	cv::Ptr<cv::aruco::DetectorParameters> parameters(new cv::aruco::DetectorParameters);
```
marip8, line 53: - What are these parameters?
- Should we expose them to the user?
- Why not use `cv::aruco::DetectorParameters::create()`?

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+	std::vector<int> marker_ids;
+    std::vector<std::vector<cv::Point2f> > marker_corners, rejected_candidates;
+    cv::aruco::detectMarkers(image, board_->dictionary, marker_corners, marker_ids, parameters);
+	cv::aruco::refineDetectedMarkers(image, board_, marker_corners, marker_ids, rejected_candidates);
+	for(int i = 0; i < marker_ids.size(); i++)
```
marip8, line 58: Since we actually want to find the chessboard corners, I think you need a call to `cv::aruco::interpolateCornersCharuco` here

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+	return map_ids_to_obs_corners;	
+}
+
+cv::Mat
+rct_image_tools::CharucoGridBoardObservationFinder::drawObservations(const cv::Mat& image,
+                                                                   const std::map<int, std::vector<Eigen::Vector2d>>& observations) const
+{
+	std::vector<int> marker_ids;
+	std::vector<std::vector<cv::Point2f>> marker_corners;
+	for(std::map<int, std::vector<Eigen::Vector2d>>::const_iterator it = observations.begin(); it != observations.end(); ++it)
```
marip8, line 78: Replace `std::map<int, std::vector<Eigen::Vector2d>>::const_iterator` with `auto` for readability

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return map_ids_to_obs_corners;	
+}
+
+cv::Mat
+rct_image_tools::CharucoGridBoardObservationFinder::drawObservations(const cv::Mat& image,
+                                                                   const std::map<int, std::vector<Eigen::Vector2d>>& observations) const
+{
+	std::vector<int> marker_ids;
+	std::vector<std::vector<cv::Point2f>> marker_corners;
```
marip8, line 77: `reserve` the size of these vectors ahead of time (`observations.size()`)

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+		marker_ids.push_back(it->first);
+		std::vector<cv::Point2f> cv_obs(it->second.size());
+		std::transform(it->second.begin(), it->second.end(), cv_obs.begin(), [](const Eigen::Vector2d& o) {return cv::Point2d(o.x(), o.y()); });
+		marker_corners.push_back(cv_obs);
+	}
+	
+	cv::aruco::drawDetectedMarkers(image, marker_corners, marker_ids);
+	std::vector<cv::Point2f> charuco_corners;
+	std::vector<int> charuco_ids;
+	cv::aruco::interpolateCornersCharuco(marker_corners, marker_ids, image, board_, charuco_corners, charuco_ids);
```
marip8, line 89: Probably don't need this code if the input observations are the ChArUco interpolated corners

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+  std::map<int, std::vector<Eigen::Vector3d>> map_ids_to_corners;
+  for (int i = 0; i < board->ids.size(); i++)
+  {
+    std::vector<Eigen::Vector3d> obj_pts(board->objPoints[i].size());
+    std::transform(board->objPoints[i].begin(), board->objPoints[i].end(), obj_pts.begin(), [](const cv::Point3f& o) {return Eigen::Vector3d(o.x, o.y, o.z); });
+    map_ids_to_corners.insert(std::make_pair(board->ids[i], obj_pts));
+  }
+  return map_ids_to_corners;
+}
```
marip8, line 105: I would refactor this function to provide a `Correspondence2D3D` type. This is what we require for our calibration problems. I think this method would become:

```c++
std::vector<Correspondence2D3D> createCorrespondences(const cv::Ptr<cv::aruco::CharucoBoard>& board)
```

rct_image_tools/include/rct_image_tools/charuco.hpp
jaberkebile: It's part of OpenCV 3.0 or greater, so is it better to copy paste it?

schornakj: ROS Kinetic and later depend on OpenCV 3.X, so as long as we're building this package on Kinetic or newer we'll always have a comparatively up-to-date version of OpenCV that includes the 3rd-party contrib packages which provide the ArUco library. This specific header is available through `#include <opencv2/aruco/charuco.hpp`.

One challenge is that the Canonical-sourced OpenCV package for Ubuntu 16.04 (Xenial) is actually OpenCV 2.X. For a potential application where `rct_image_tools` is being used in a non-ROS context on a Xenial system, we'd have to specifically require OpenCV 3 in our `CMakeLists.txt`. This would require the user to manually provide that version of the library rather than being able to rely on `apt install` to provide the right version. I don't think this is a use case we need to specifically support, since as currently designed the higher-level packages in this repo like `rct_examples` depend on ROS (and by the time we'll probably get around to making the whole repo ROS-agnostic it'll be April 2021 and Xenial will finally be out of support).

As a side node OpenCV's file naming conventions are kind of weird. I think they did a big refactor between 1.X and 2.X and changed the directory of their headers to use `opencv2`. However, 3.X uses the same directory, so there isn't an `opencv3` directory.

rct_image_tools/src/rct_image_tools/charuco_finder.cpp
schornakj: In `aruco_finder.cpp` the purpose of this function is to generate a map between marker IDs and expected target-relative points, which enables lookup of the target points for each visible marker ID. Changing the function to return a vector of correspondences makes it do something fundamentally different than what was intended.

@jaberkebile I think that the way ChArUco grids are detected and represented might be sufficiently different from regular ArUco grids that you might have to fundamentally revisit how you detect observed image points and correlate them to possible target object points. If you can write up a unit test that loads in pictures of ChArUco boards (with some partial views) and uses the detected correlations to perform an extrinsic hand-eye optimization, that should expose these kinds of differences and guide the development of a solution.

marip8:
`rct_image_tools` is a ROS-independent, pure CMake package such that it can support usage in ROS1 and ROS2. Please remove the dependency on `catkin` and ROS1 from the latest commit. You should be able specify local file directories via an interface target in the `CMakeLists.txt`. 

Here is an example from `tesseract`:
- [Specification of interface target](https://github.com/ros-industrial-consortium/tesseract/blob/b9284bd16bfd34f43a2b79509471ef386ffb4d02/tesseract/tesseract_support/CMakeLists.txt) where the directory [`TESSERACT_SUPPORT_DIR` is defined as a variable](https://github.com/ros-industrial-consortium/tesseract/blob/b9284bd16bfd34f43a2b79509471ef386ffb4d02/tesseract/tesseract_support/CMakeLists.txt#L8).
- [Link your executable against this interface target](https://github.com/ros-industrial-consortium/tesseract/blob/b9284bd16bfd34f43a2b79509471ef386ffb4d02/tesseract/tesseract_planning/tesseract_motion_planners/test/CMakeLists.txt#L6)
- [Usage of the variable](https://github.com/ros-industrial-consortium/tesseract/blob/b9284bd16bfd34f43a2b79509471ef386ffb4d02/tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp#L156) provided by the interface target

marip8:
It also appears that the build is failing on Xenial

```bash
  /root/target_ws/src/robot_cal_tools/rct_image_tools/src/rct_image_tools/charuco_finder.cpp:15:17: error: ‘struct cv::aruco::DetectorParameters’ has no member named ‘doCornerRefinement’
  
       parameters->doCornerRefinement = false;
```

PR #78, opened 2020-08-14
Circle Detector Update

marip8:
This PR modifies the circle detector to improve robustness and add debug capability. The main changes in this PR are:
1. General clean up of the circle detector code
1. Bug fix for incorrect determination of features that were the same in differently threshold-ed images
1. Addition of contour to ellipse model check that more effectively filters non-circular points
1. Addition of debug flag that shows the found circles displayed on the threshold-ed image

### DIFF
--- a/rct_image_tools/src/rct_image_tools/circle_detector.cpp
+++ b/rct_image_tools/src/rct_image_tools/circle_detector.cpp
@@ -51,80 +51,28 @@ the use of this software, even if advised of the possibility of such damage.
 #include <algorithm>
 #include <iterator>
 #include <yaml-cpp/yaml.h>
-#include <opencv2/highgui.hpp>
 
-namespace rct_image_tools
+namespace
 {
-CircleDetector::Params::Params()
+/**
+ * @brief The DetectionOutput struct
+ */
+struct DetectionResult
 {
-  minThreshold = 50;
-  maxThreshold = 220;
-  nThresholds = 18;
-
-  minRepeatability = 3;
-  circleInclusionRadius = 5;
-  maxRadiusDiff = 5;
-
-  maxAverageEllipseError = 0.02;
-
-  filterByColor = true;
-  circleColor = 0;
-
-  filterByArea = true;
-  minArea = 25;
-  maxArea = 5000;
-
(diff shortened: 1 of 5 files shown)


rct_image_tools/src/rct_image_tools/circle_detector.cpp
```diff
@@ -315,12 +427,12 @@ bool CircleDetector::loadParams(const std::string& path, CircleDetector::Params&
 
   try
   {
-    optionalLoad(n, "thresholdStep", p.thresholdStep);
+    optionalLoad(n, "thresholdStep", p.nThresholds);
```
jdlangs, line 430: I'm guessing the three parameter names here should change?

rct_image_tools/src/rct_image_tools/image_observation_finder.cpp
```diff
@@ -321,15 +321,15 @@ static bool extractKeyPoints(const ObservationPoints& centers, ObservationPoints
 
 template <typename PARAMS, typename DETECTOR_PTR, typename DETECTOR>
 static bool extractModifiedCircleGrid(const cv::Mat& image, const rct_image_tools::ModifiedCircleGridTarget& target,
-                                      ObservationPoints& observation_points, const PARAMS* set_params = nullptr)
+                                      ObservationPoints& observation_points, const bool debug, const PARAMS* set_params = nullptr)
```
jdlangs, line 324: I'm not a fan of adding a boolean flag to the object constructors to control their internal behavior. If feasible, I think it would be better to expose the debug image creation/display as a standalone function applications could call. That way you aren't exposing it to all users who will see the flag in the declaration and wonder what it does.

marip8: I did a bit of an overhaul to the `CircleDetector` class. Let me know what you think

marip8:
@jdlangs any further thoughts about this PR?

jdlangs:
Thanks for reminding me about this. The new API looks good and makes a lot of sense. Nicely done.

marip8:
Would you mind approving if there aren't any other changes that need to be made?

PR #74, opened 2020-07-27
Add kinematic optimization using 6-DOF pose observations

schornakj:
Depends on #72 -- needs to be revised after that one merges.

Add a cost function and optimization that uses the `KinematicMeasurement` class.

PR #79, opened 2020-08-25
Kinematic Optimization with 6D Pose Measurements

marip8:
This PR supercedes #74. Notable changes include:
- Changed residual calculation to include x/y/z difference and a scaled quaternion angular difference
- Removed unit test residual error checks because of the mixed linear and rotational units in the residuals
- Added an example with data collected on a robotic part positioner


marip8:
# Results from the example

## Calibration validation against measurements:
Position Difference Mean: 0.00174474
Position Difference Std. Dev.: 0.000923751
Orientation Difference Mean: 0.00323072
Orientation difference Std. Dev.: 0.00175761

## Calibration validation against measurements (static DH parameters):
Position Difference Mean: 0.00382386
Position Difference Std. Dev.: 0.00144508
Orientation Difference Mean: 0.00534404
Orientation difference Std. Dev.: 0.00219113

## Percent improvement: calibration vs. nominal kinematic model
Position: 54.3723%
Position Std. Dev.: 36.0761%
Orientation: 39.5453%
Orientation Std. Dev.: 19.7856%

rct_optimizations/src/rct_optimizations/dh_chain_kinematic_calibration.cpp
```diff
@@ -10,6 +10,7 @@
 namespace rct_optimizations
 {
 
+// TODO: replace; duplicates a member function
```
schornakj, line 13: I think we can delete this comment. [The function I think I was referring to](https://github.com/Jmeyer1292/robot_cal_tools/blob/659c0756affde260c842c8f20d20598903a27bb6/rct_optimizations/include/rct_optimizations/dh_chain_kinematic_calibration.h#L149-L163) when I added the comment ended up being substantially different from the one here.

PR #86, opened 2020-10-19
Add explicit Boost dependency

schornakj:
Depends on #85 

ROS2 no longer depends on Boost, since many of the features that ROS1 was using like `boost::shared_ptr` have been integrated into the C++ standard library. Our `rct_image_tools` package uses some Boost features, specifically `boost::optional`, so it fails to build in ROS2.

A quick fix is to explicitly depend on, find, and link against Boost in `rct_image_tools`.

A longer-term fix would be to minimize our use of Boost in general. There is now a `std::optional` in C++17, but it might be better to redesign the functions that return optionals to throw exceptions instead.

### DIFF
--- a/rct_image_tools/CMakeLists.txt
+++ b/rct_image_tools/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.5.0)
 project(rct_image_tools VERSION 0.1.0 LANGUAGES CXX)
 
+find_package(Boost REQUIRED)
 find_package(OpenCV REQUIRED)
 find_package(yaml-cpp REQUIRED)
 find_package(rct_common REQUIRED)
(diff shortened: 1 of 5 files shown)


jdlangs:
:+1:  to not using optionals for error-handling purposes. But we might as well do the switch to `std::optional` and C++17 since it's supported by the default gcc even in 18.04. If that's all that's needed to avoid a Boost dependency, that is.

schornakj:
> +1 to not using optionals for error-handling purposes. But we might as well do the switch to `std::optional` and C++17 since it's supported by the default gcc even in 18.04. If that's all that's needed to avoid a Boost dependency, that is.

I remembered that `rct_optimizations` uses Boost accumulators ([for example, here](https://github.com/Jmeyer1292/robot_cal_tools/blob/d0cd5428d53fbc24c65a7c302631a875d50bea99/rct_optimizations/src/rct_optimizations/validation/noise_qualification.cpp)), which as far as I know can't be directly replaced with a standard-library equivalent. I think I'll need to add my dependency modification to `rct_optimizations` as well.

marip8:
I think we should eventually just remove the functions that use `boost::optional` and prefer the ones that throw exceptions.

schornakj:
I added explicit Boost dependencies to `rct_optimizations` as well for now. @marip8, this is ready for review once CI passes.

> I think we should eventually just remove the functions that use `boost::optional` and prefer the ones that throw exceptions.

I'll make this modification in an upcoming PR.

marip8 (approved):

PR #87, opened 2020-10-30
Update Target Finder Interface

marip8:
This PR creates a common target finder interface to allow for applications to easily switch between different types of calibration targets.


marip8:
@schornakj can you take a look at this?

PR #91, opened 2020-11-21
Circle Detector Bug Fix

marip8:
This PR updates the custom circle finder to remove contours with less than 5 points. OpenCV throws an exception when trying to fit an ellipse to less than 5 points

### DIFF
--- a/rct_image_tools/src/rct_image_tools/circle_detector.cpp
+++ b/rct_image_tools/src/rct_image_tools/circle_detector.cpp
@@ -82,6 +82,12 @@ DetectionResult findCircles(const cv::Mat& image, const double threshold,
   std::vector<std::vector<cv::Point>> contours;
   cv::findContours(binarized_image, contours, cv::RETR_LIST, cv::CHAIN_APPROX_NONE);
 
+  // Remove all contours with less than 5 points (OpenCV requirement for ellipse fitting)
+  contours.erase(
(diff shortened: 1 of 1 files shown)


schornakj:
@marip8 This looks good to me. As a note for later work, we should add an image to the circle-finder unit tests that's really borderline to expose these kinds of exceptions.

Since I first learned about `std::remove_if` I've found it a little funny that it doesn't actually remove anything itself...

marip8:
> As a note for later work, we should add an image to the circle-finder unit tests that's really borderline to expose these kinds of exceptions.

I had not actually encountered this issue until last week. OpenCV found a whole bunch of contours with lengths of zero to 4 in several of the images that I was working with. I can upload one of them and create a test if it's valuable. What did you have in mind for the unit test, now that this issue has been addressed?

schornakj:
> 
> 
> > As a note for later work, we should add an image to the circle-finder unit tests that's really borderline to expose these kinds of exceptions.
> 
> I had not actually encountered this issue until last week. OpenCV found a whole bunch of contours with lengths of zero to 4 in several of the images that I was working with. I can upload one of them and create a test if it's valuable. What did you have in mind for the unit test, now that this issue has been addressed?

I was thinking of including some images that have caused bad results in the past to make sure that we don't reintroduce the various bugs we've fixed. Another example could be an image of a target where the circle centers are returned in the wrong order.

PR #92, opened 2020-11-21
Update Serialization

marip8:
This PR adds serialization to some of the types defined in `rct_optimizations`

### DIFF
--- a/rct_optimizations/CMakeLists.txt
+++ b/rct_optimizations/CMakeLists.txt
@@ -4,6 +4,7 @@ project(rct_optimizations VERSION 0.1.0 LANGUAGES CXX)
 find_package(rct_common REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Ceres REQUIRED)
+find_package(yaml-cpp REQUIRED)
 
 add_library(${PROJECT_NAME} SHARED
   # Utilities
@@ -45,6 +46,7 @@ target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC
 target_link_libraries(${PROJECT_NAME} PUBLIC
   ${Boost_LIBRARIES}
   ${CERES_LIBRARIES}
+  yaml-cpp
 )
 
 if(RCT_BUILD_TESTS)
--- a/rct_optimizations/include/rct_optimizations/serialization/eigen.h
+++ b/rct_optimizations/include/rct_optimizations/serialization/eigen.h
@@ -5,10 +5,10 @@
 
 namespace YAML
 {
-template<typename V>
-struct convert<Eigen::Matrix<V, 2, 1>>
+template<typename FloatT>
+struct convert<Eigen::Matrix<FloatT, 2, 1>>
 {
-  using T = Eigen::Matrix<V, 2, 1>;
+  using T = Eigen::Matrix<FloatT, 2, 1>;
 
   static Node encode(const T &val)
   {
@@ -23,17 +23,17 @@ struct convert<Eigen::Matrix<V, 2, 1>>
     if (node.size() != 2)
       return false;
 
-    val.x() = node["x"].as<V>();
-    val.y() = node["y"].as<V>();
+    val.x() = node["x"].as<FloatT>();
+    val.y() = node["y"].as<FloatT>();
 
     return true;
   }
 };
 
-template<typename V>
-struct convert<Eigen::Matrix<V, 3, 1>>
+template<typename FloatT>
+struct convert<Eigen::Matrix<FloatT, 3, 1>>
 {
-  using T = Eigen::Matrix<V, 3, 1>;
+  using T = Eigen::Matrix<FloatT, 3, 1>;
 
   static Node encode(const T &val)
   {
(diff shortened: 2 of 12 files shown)


marip8:
@schornakj can you review?

rct_optimizations/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(rct_optimizations VERSION 0.1.0 LANGUAGES CXX)
 find_package(rct_common REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Ceres REQUIRED)
+find_package(yaml-cpp REQUIRED)
```
schornakj, line 7: Should also be added as a dependency in `package.xml`.

rct_optimizations/include/rct_optimizations/serialization/eigen.h
```diff
@@ -0,0 +1,103 @@
+#pragma once
+
+#include <Eigen/Geometry>
+#include <yaml-cpp/yaml.h>
+
+namespace YAML
+{
+template<typename V>
+struct convert<Eigen::Matrix<V, 2, 1>>
```
schornakj, line 9: Is there any possibility of these functions colliding with their equivalents in [`message_serialization`](https://github.com/swri-robotics/message_serialization/blob/master/include/message_serialization/eigen_yaml.h) if both are included in the same file? I made a very simple test case to check for this and it didn't immediately cause problems, but I don't know if there's some other subtle issue to watch out for.

schornakj (changes requested):
@marip8 Looks good in general, just a few questions.

I would also suggest adding some unit tests to at minimum verify that converting various types to YAML and then back again returns the same original value.

rct_optimizations/include/rct_optimizations/serialization/eigen.h
marip8: I'm not entirely sure; I haven't experienced any problems with it either yet. I think this is something we can figure out in the future if it happens though

drchrislewis: Why don't we just add these to message_serialization instead of here. These data types are not specific to robot calibration in any way.

rct_optimizations/include/rct_optimizations/serialization/problems.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
```
drchrislewis, line 1: These structures on the other hand are specific to this repo and belong here.

rct_optimizations/include/rct_optimizations/serialization/problems.h
```diff
@@ -0,0 +1,88 @@
...
+
+  static Node encode(const T &rhs)
+  {
+    Node node;
+
+    node["intr"] = rhs.intr;
+    node["camera_mount_to_camera_guess"] = rhs.camera_mount_to_camera_guess;
+    node["target_mount_to_target_guess"] = rhs.target_mount_to_target_guess;
+    node["observations"] = rhs.observations;
+
```
drchrislewis, line 21: There are two non-constant string labels label_target_mount_to_target and label_camera_mount_to_camera that may have been intended to be constant values, but are neither read nor written.

rct_optimizations/include/rct_optimizations/serialization/problems.h
```diff
@@ -0,0 +1,88 @@
...
+    node["target_mount_to_target_pos"] = rhs.target_mount_to_target;
+    node["camera_mount_to_camera_pos"] = rhs.camera_mount_to_camera;
+
+    // Serialize transform orientation in RPY for convenience
+    node["target_mount_to_target_rpy"] = Eigen::Vector3d (rhs.target_mount_to_target.rotation().eulerAngles(2, 1, 0).reverse());
+    node["camera_mount_to_camera_rpy"] = Eigen::Vector3d(rhs.camera_mount_to_camera.rotation().eulerAngles(2, 1, 0).reverse());
+    return node;
+  }
+};
+
```
drchrislewis, line 87: decode is missing for ExtrinsicEyeHandResult

drchrislewis (approved):

drchrislewis:
@marip8 @schornakj I made some suggestions, but nothing major. I approved, but did not merge.

rct_optimizations/include/rct_optimizations/serialization/eigen.h
marip8: Most, if not all, of these functions are already in `message_serialization`. I avoided adding a dependency on `message_serialization` here because it is currently a ROS1 package and `rct_optimizations` is a pure CMake package. At this point I don't mind duplicating a small number of serialization functions if it means not adding an external build-from-source dependency.

rct_optimizations/include/rct_optimizations/serialization/problems.h
marip8: Those labels are used for printing covariance information, so I didn't think it was necessary to include them in the serialization. I can add them for the sake of completeness though.

Those labels are non-const so that the user can change them if they want to; that was an intentional design choice I believe.

rct_optimizations/include/rct_optimizations/serialization/problems.h
marip8: I'll add it

PR #93, opened 2020-12-09
Add alternative constructor for CharucoGridTarget

schornakj:
Similar to #90.

Add a constructor for`CharucoGridTarget` that takes a `cv::Ptr<cv::aruco::GridBoard>` as an argument, to allow using externally-defined gridboard objects.

### DIFF
--- a/rct_image_tools/src/rct_image_tools/charuco_grid_target.cpp
+++ b/rct_image_tools/src/rct_image_tools/charuco_grid_target.cpp
@@ -1,22 +1,37 @@
 #include <rct_image_tools/charuco_grid_target.h>
 
-namespace rct_image_tools
+namespace
 {
-CharucoGridTarget::CharucoGridTarget(const int rows, const int cols, const float chessboard_dim,
-                                     const float aruco_marker_dim, const int dictionary_id)
+/**
+ * @brief For a given ChArUco board, create a map of chessboard intersection coordinates keyed to marker IDs.
+ * @param board - ChArUco board to use when generating the map
+ * @return Resulting map. Keys are the indices of chessboard intersections in the board. Values are board-relative coordinates for each intersection.
+ */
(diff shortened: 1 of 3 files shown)


rct_image_tools/src/rct_image_tools/charuco_grid_target.cpp
```diff
@@ -1,22 +1,37 @@
...
+namespace
 {
-CharucoGridTarget::CharucoGridTarget(const int rows, const int cols, const float chessboard_dim,
-                                     const float aruco_marker_dim, const int dictionary_id)
+/**
+ * @brief For a given ChArUco board, create a map of chessboard intersection coordinates keyed to marker IDs.
+ * @param board - ChArUco board to use when generating the map
+ * @return Resulting map. Keys are the indices of chessboard intersections in the board. Values are board-relative coordinates for each intersection.
+ */
+std::map<unsigned, Eigen::Vector3d> mapCharucoIdsToObjPts(const cv::Ptr<cv::aruco::CharucoBoard> &board)
```
marip8, line 10:
```suggestion
std::map<unsigned, Eigen::Vector3d> mapCharucoIdsToObjPts(const cv::Ptr<const cv::aruco::CharucoBoard> &board)
```
Would it be possible to use a `const` pointer here? I think that would be preferred since we don't need to modify the board object

schornakj: This change seems to cause some problems with OpenCV's ArUco-related functions, which are all written to take this argument as `const cv::Ptr<CharucoBoard> &board`. I think the issue is that a `const cv::Ptr<CharucoBoard> &board` is not interchangeable with a `const cv::Ptr<const CharucoBoard> &board`.

marip8 (approved):

PR #96, opened 2021-07-20
Target Finder Plugins

marip8:
This PR adds the ability to specify the type of target to be used for a calibration via plugin at run-time without the need to recompile. Specifically this PR:
- Adds a plugin interface in `rct_ros_tools` for dynamically loading different types of target finders (e.g. modified circle grid, ChArUco)
- Revises tools and examples in `rct_examples` to utilize `pluginlib` to dynamically load a target finder from ROS parameter
- Various small revisions to the tools/examples to print more useful data about the calibrations and associated analysis


marip8:
@jdlangs friendly ping for review

rct_ros_tools/include/rct_ros_tools/target_finder_plugin.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <ros/node_handle.h>
```
jdlangs, line 3: Looks like this can be removed?

rct_ros_tools/include/rct_ros_tools/target_finder_plugin.h
```diff
@@ -0,0 +1,41 @@
...
+    return finder_->drawTargetFeatures(image, target_features);
+  }
+
+  const rct_image_tools::Target& target() const override
+  {
+    return finder_->target();
+  }
+
+  virtual void init(const XmlRpc::XmlRpcValue& config) = 0;
+  virtual void init(const std::string& file) = 0;
```
jdlangs, line 35: Might be better to drop the XmlRpc option and only use yaml which would make this easier to move to ROS2. A simple XmlRpc -> yaml converter could be provided for the parameter server use case.

jdlangs (approved):
I didn't thoroughly review all the changes but I looked at all the plugin stuff and everything looks good and makes sense.

rct_ros_tools/include/rct_ros_tools/target_finder_plugin.h
marip8: So maybe we unify on `YAML::Node` instead?

jdlangs: That seems reasonable to me.

PR #97, opened 2022-02-02
Devel/kinematic calibration

Levi-Armstrong:
@marip8 I noticed that you had this branch wanted to check and see if it would still be good to get merged? I know a while a go some work was done to refactor things into industrial_calibration, but not sure what the status is.


marip8:
I'm okay with merging the optimization changes. I don't think the tool is needed; it's pretty specific to a single use-case and there is already an example (possibly the same file with some minor changes) in the repository

marip8:
> I know a while a go some work was done to refactor things into industrial_calibration, but not sure what the status is.

Hasn't progressed much; I'm okay with adding here and porting to industrial_calibration later when we focus on updating it

PR #99, opened 2022-02-11
Fix dh parameter labels and add method for calculating covariance in tangent space

Levi-Armstrong:
This addresses two things and exposes one issue ~~and not sure how to fix~~.
- The parameter labels in stored in the optimize function were not correct because Eigen is column major so the lables need to be stored the same way.
- The compute covariance includes all parameters even if a parameter block is constant or sub parameters are constant. This adds another function but was thinking about updating the existing to do this. @marip8 Let me know what you think about updating the existing functions to do this instead of having a separate function?
- Also noticed something odd where on some tests the parameter block was in an incorrect order. I have added a check and now the unit tests are failing because this is being detected. Some how the target dh parameter block and camera dh parameter block is being swapped. @marip8 Do you know why this would be happening.

### DIFF
--- a/rct_optimizations/include/rct_optimizations/covariance_analysis.h
+++ b/rct_optimizations/include/rct_optimizations/covariance_analysis.h
@@ -38,7 +38,6 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
 
 /**
  * @brief Compute all covariance results for a Ceres optimization problem. Labels results with generic names.
- * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
  * @param options ceres::Covariance::Options to use when calculating covariance.
@@ -50,7 +49,6 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
 
 /**
  * @brief Compute covariance results for the specified parameter blocks in a Ceres optimization problem. Labels results with generic names.
- * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_blocks Specific parameter blocks to compute covariance between.
  * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
@@ -64,7 +62,6 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
 
 /**
  * @brief Compute all covariance results for a Ceres optimization problem and label them with the provided names.
- * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_names Labels for all optimization parameters in the problem.
  * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
@@ -78,7 +75,6 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
 
 /**
  * @brief Compute covariance results for specified parameter blocks in a Ceres optimization problem and label them with the provided names.
- * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_blocks Specific parameter blocks for which covariance will be calculated.
  * @param parameter_names Labels for optimization parameters in the specified blocks.
--- a/rct_optimizations/include/rct_optimizations/covariance_analysis.h
+++ b/rct_optimizations/include/rct_optimizations/covariance_analysis.h
@@ -21,6 +21,14 @@ struct DefaultCovarianceOptions : ceres::Covariance::Options
 };
 
 /**
+ * @brief Get the covariance results provided covariance matrix and parameter names
+ * @param cov_matrix The covariance matrix
+ * @param parameter_names The parameter names
+ * @return The covariance results
+ */
+CovarianceResult computeCovarianceResults(const Eigen::MatrixXd& cov_matrix, const std::vector<std::string>& parameter_names);
+
+/**
  * @brief Given a covariance matrix, calculate the matrix of standard deviations and correlation coefficients.
  * @param Covariance matrix
  * @return Matrix of standard deviations (diagonal elements) and correlation coefficents (off-diagonal elements).
@@ -30,40 +38,51 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
 
 /**
  * @brief Compute all covariance results for a Ceres optimization problem. Labels results with generic names.
+ * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
+ * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
  * @param options ceres::Covariance::Options to use when calculating covariance.
  * @return CovarianceResult for the problem.
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
 
 /**
  * @brief Compute covariance results for the specified parameter blocks in a Ceres optimization problem. Labels results with generic names.
+ * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_blocks Specific parameter blocks to compute covariance between.
+ * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
  * @param options ceres::Covariance::Options to use when calculating covariance.
  * @return CovarianceResult for the problem.
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
                                    const std::vector<const double *>& parameter_blocks,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
 
 /**
  * @brief Compute all covariance results for a Ceres optimization problem and label them with the provided names.
+ * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_names Labels for all optimization parameters in the problem.
+ * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
  * @param options ceres::Covariance::Options to use when calculating covariance.
  * @return CovarianceResult for the problem.
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
-                                   const std::vector<std::vector<std::string>>& parameter_names,
+                                   const std::map<const double *, std::vector<std::string> > &param_names,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
 
 /**
  * @brief Compute covariance results for specified parameter blocks in a Ceres optimization problem and label them with the provided names.
+ * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_blocks Specific parameter blocks for which covariance will be calculated.
  * @param parameter_names Labels for optimization parameters in the specified blocks.
+ * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
  * @param options ceres::Covariance::Options to use when calculating covariance.
  * @return CovarianceResult for the problem.
  * @throw CovarianceException if covariance.Compute fails.
@@ -73,6 +92,8 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
                                    const std::vector<const double *>& parameter_blocks,
-                                   const std::vector<std::vector<std::string>>& parameter_names,
+                                   const std::map<const double *, std::vector<std::string> > &param_names,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
+
 }  // namespace rct_optimizations
--- a/rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
+++ b/rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
@@ -36,7 +36,57 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
   return out;
 }
 
+CovarianceResult computeCovarianceResults(const Eigen::MatrixXd& cov_matrix, const std::vector<std::string>& parameter_names)
+{
+  // 0. Save original covariance matrix output
+  // For parameter blocks [p1, p2], the structure of this matrix will be:
+  /*    |     p1    |     p2    |
+   * ---|-----------|-----------|
+   * p1 | C(p1, p1) | C(p1, p2) |
+   * p2 | C(p2, p1) | C(p2, p2) |
+   */
+  CovarianceResult res;
+  res.covariance_matrix = cov_matrix;
+
+  // 1. Compute matrix of standard deviations and correlation coefficients.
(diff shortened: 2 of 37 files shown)


Levi-Armstrong:
It turns out that some the code assumes the vector of parameter blocks returned from ceres would be in the same order added. This is not the case because ceres is store it in a map and where requested it iterates over the map and create a vector using the key so the order may not be the same.

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -171,37 +171,40 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
...
-    int block_size = problem.ParameterBlockSize(parameter_blocks[index_block]);
-    if (static_cast<std::size_t>(block_size) != parameter_names[index_block].size())
+    int block_size = problem.ParameterBlockSize(b);
+    if (static_cast<std::size_t>(block_size) != param_names.at(b).size())
     {
       std::stringstream ss;
-      ss << "Number of parameter labels provided for block " << index_block << " does not match actual number of parameters in that block: " \
-         << "have " << parameter_names[index_block].size() << " labels and " << block_size << " parameters";
+      ss << "Number of parameter labels provided for block does not match actual number of parameters in that block: " \
+         << "have " << param_names.at(b).size() << " labels and " << block_size << " parameters";
```
marip8, line 199: Add the name of the block in the error message

```suggestion
      ss << "Number of parameter labels provided for block '" << param_names.at(b).first << "' does not match actual number of parameters in that block: " \
         << "have " << param_names.at(b).size() << " labels and " << block_size << " parameters";
```

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -169,4 +169,118 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
 
   return res;
 }
+
+CovarianceResult computeCovarianceInTangentSpace(ceres::Problem &problem,
```
marip8, line 173: Did you look at using [this function in Ceres](http://ceres-solver.org/nnls_covariance.html#_CPPv4NK5ceres32GetCovarianceBlockInTangentSpaceEPKdPKdPd) to get the covariance block of the tangent space? Seems like it might be equivalent and much easier than adding a whole new function

marip8 (changes requested):
Overall I'm on board with this change. I think it's worth looking into a simpler way extracting the covariance of the tangent space though

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
Levi-Armstrong: I tried that first but was unsuccessful due to runtime crashing. I assumed it was because the size of the covariance matrix being passed in was not correct, which led me to to believe it was including other parameters that I wanted to ignore so I ended up with this.

Levi-Armstrong:
@marip8 Do you think all the function should be updated to return this as the covariance or keep it as a separate?

Levi-Armstrong:
@marip8 you good with me merging this?

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -116,57 +169,81 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
...
 
-  // 5. Create NamedParams for covariance and correlation results, which include labels and values for the parameters. Uses top-right triangular part of matrix.
-  Eigen::Index col_start = 0;
-  for (Eigen::Index row = 0; row < correlation_matrix.rows(); row++)
+  if (parameter_blocks.size() != param_masks.size())
+    throw CovarianceException("Provided vector param_masks is not same length as provided number of parameter blocks");
+
+  Eigen::Index n_params_in_selected = 0;
+  std::vector<Eigen::Index> tagent_space_indices;
+  std::vector<std::string> tagent_space_labels;
```
marip8, line 194:
```suggestion
  std::vector<Eigen::Index> tangent_space_indices;
  std::vector<std::string> tangent_space_labels;
```

rct_optimizations/src/rct_optimizations/dh_chain_kinematic_calibration.cpp
```diff
@@ -416,6 +480,9 @@ KinematicCalibrationResult optimize(const KinematicCalibrationProblemPose6D &par
     problem.AddResidualBlock(cost_block, nullptr, target_chain_dh_offsets.data());
   }
 
+  // Print optimization parameter labels
+  printOptimizationLabels(problem, param_names, param_labels, param_masks);
+
```
marip8, line 485: Is this really needed or is this for debug?

marip8 (changes requested):
I haven't had a chance to fully look through this yet; I'll look at it in more detail tonight. One thought though: does it ever make sense not to compute the covariance in the tangent space? It seems like covariance between static variables is not valid

Levi-Armstrong:
> I haven't had a chance to fully look through this yet; I'll look at it in more detail tonight. One thought though: does it ever make sense not to compute the covariance in the tangent space? It seems like covariance between static variables is not valid

Yea, that was my thought is it is not needed. If you wanted the full covariance you would just pass it an empty masks. If you onboard I can just remove the new function and update the documentation on the other functions. Let me know if you agree and I will make the change.

rct_optimizations/src/rct_optimizations/dh_chain_kinematic_calibration.cpp
Levi-Armstrong: Just for debug.

marip8:
> If you onboard I can just remove the new function and update the documentation on the other functions. Let me know if you agree and I will make the change.

I think that makes sense; I'm on board with doing that. Like you mentioned, if you want the full covariance you can just pass empty masks (or we could make that the default value for that argument)

rct_common/include/rct_common/extrinsic_data_set.h
```diff
@@ -0,0 +1,4 @@
+#ifndef EXTRINSIC_DATA_SET_H
+#define EXTRINSIC_DATA_SET_H
+
+#endif // EXTRINSIC_DATA_SET_H
```
marip8, line 4: Is this necessary?

Levi-Armstrong: Nope

rct_optimizations/include/rct_optimizations/covariance_analysis.h
```diff
@@ -30,40 +38,48 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
...
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
 
 /**
  * @brief Compute covariance results for specified parameter blocks in a Ceres optimization problem and label them with the provided names.
+ * @details This excludes all parameters set to be constant during the solve
  * @param problem The Ceres problem (after optimization).
  * @param parameter_blocks Specific parameter blocks for which covariance will be calculated.
  * @param parameter_names Labels for optimization parameters in the specified blocks.
+ * @param param_masks The mask parameters associated with parameter block
```
marip8, line 82:
```suggestion
 * @param param_masks Map of the parameter block pointer and the indices of the parameters within that block to be excluded from the covariance calculation
```

rct_optimizations/include/rct_optimizations/covariance_analysis.h
```diff
@@ -73,6 +89,8 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
                                    const std::vector<const double *>& parameter_blocks,
-                                   const std::vector<std::vector<std::string>>& parameter_names,
+                                   const std::map<const double *, std::vector<std::string> > &param_names,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
```
marip8, line 93:
```suggestion
                                   const std::map<const double*, std::vector<std::string>> &param_names,
                                   const std::map<const double*, std::vector<int>>& param_masks = std::map<const double*, std::vector<int>>(),
```

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -36,22 +36,73 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
...
+      res.covariances.push_back(p_cov);
+    }
+    col_start++;
+  }
+
+  return res;
+}
+
 CovarianceResult computeCovariance(ceres::Problem &problem,
+                                   const std::map<const double *, std::vector<int> > &param_masks,
```
marip8, line 89:
```suggestion
                                   const std::map<const double*, std::vector<int>> &param_masks,
```

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -36,22 +36,73 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
...
+  return computeCovariance(problem, blocks_const, param_masks, options);
 }
 
 CovarianceResult computeCovariance(ceres::Problem &problem,
                                    const std::vector<const double *>& parameter_blocks,
+                                   const std::map<const double *, std::vector<int> > &param_masks,
                                    const ceres::Covariance::Options& options)
 {
-  std::vector<std::vector<std::string>> dummy_param_names;
+  std::map<const double *, std::vector<std::string>> dummy_param_names;
```
marip8, line 105:
```suggestion
  std::map<const double*, std::vector<std::string>> dummy_param_names;
```

rct_optimizations/src/rct_optimizations/covariance_analysis.cpp
```diff
@@ -61,15 +112,16 @@ CovarianceResult computeCovariance(ceres::Problem &problem,
...
 
-  return computeCovariance(problem, parameter_blocks, dummy_param_names, options);
+  return computeCovariance(problem, parameter_blocks, dummy_param_names, param_masks, options);
 
 }
 
 CovarianceResult computeCovariance(ceres::Problem &problem,
-                                   const std::vector<std::vector<std::string>>& parameter_names,
+                                   const std::map<const double *, std::vector<std::string> > &param_names,
+                                   const std::map<const double *, std::vector<int> > &param_masks,
```
marip8, line 124:
```suggestion
                                   const std::map<const double*, std::vector<std::string>> &param_names,
                                   const std::map<const double*, std::vector<int>> &param_masks,
```

rct_ros_tools/include/rct_ros_tools/print_utils.h
```diff
@@ -1,144 +1,21 @@
...
+using rct_common::getStringQuaternion;
+using rct_common::getStringRPY;
+using rct_common::getStringXYZ;
+using rct_common::printCameraDistortion;
+using rct_common::printCameraIntrinsics;
+using rct_common::printNewLine;
+using rct_common::printOptResults;
+using rct_common::printTitle;
+using rct_common::printTransform;
+using rct_common::printTransformDiff;
```
marip8, line 19: This seems unnecessary; let's get rid of this file and just change references to this function to use the `rct_common` namespace

Levi-Armstrong: This was to prevent breaking changes, but I am fine without removing.

rct_optimizations/include/rct_optimizations/covariance_analysis.h
```diff
@@ -30,40 +38,48 @@ Eigen::MatrixXd computeCorrelationsFromCovariance(const Eigen::MatrixXd& covaria
...
  */
 CovarianceResult computeCovariance(ceres::Problem &problem,
-                                   const std::vector<std::vector<std::string>>& parameter_names,
+                                   const std::map<const double *, std::vector<std::string> > &param_names,
+                                   const std::map<const double *, std::vector<int> >& param_masks = std::map<const double *, std::vector<int>>(),
                                    const ceres::Covariance::Options& options = DefaultCovarianceOptions());
 
 /**
  * @brief Compute covariance results for specified parameter blocks in a Ceres optimization problem and label them with the provided names.
+ * @details This excludes all parameters set to be constant during the solve
```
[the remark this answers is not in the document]
Levi-Armstrong: This is a valid comment. This does occur regardless of the mask now.

Levi-Armstrong:
> It seems like this PR is making 5 sets of unrelated changes:
> 
> * Removing the Xenial build
> * Fix for YAML include
> * Moving print headers to `rct_common`
> * Adding a helper to the DH chain to get a relative transform
> * Updating the covariance calculation
> 
> I'm okay with making these changes, but I would really prefer if we merged them as separate PRs in the order above. It makes it hard to look back and understand where changes were introduced if too many unrelated ones are made in the same PR

I wound not agree these are unrelated changes. These were all necessary to get something that was broken tested and fixed. I will clean up the commit history but I am not going to separate them into different PRs. In the end the only thing that is important is that the commit history reflect these changes.

Levi-Armstrong:
I will follow up with a different PR to add a clang format file, CI Job and format the entire repo.

rct_optimizations/include/rct_optimizations/covariance_analysis.h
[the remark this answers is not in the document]
marip8: Okay; I assume this is because we are checking `isParameterBlockConstant` in the function?

marip8:
> I will clean up the commit history but I am not going to separate them into different PRs. In the end the only thing that is important is that the commit history reflect these changes.

That's fine; the main thing is being able to see the individual changes

Levi-Armstrong:
> > I will clean up the commit history but I am not going to separate them into different PRs. In the end the only thing that is important is that the commit history reflect these changes.
> 
> That's fine; the main thing is being able to see the individual changes

Sorry, I reread my comment. Please don't take it the wrong way. I understand your point but in this case I needed all of these changes with the exception of the moving the print utils in one branch to for everything to work on a machine. The CI change was also need because some change I made broke the xenial build and I did not want to spend time fixing it for something that was EOL.

rct_ros_tools/include/rct_ros_tools/print_utils.h
Levi-Armstrong: I removed this header and updated everywhere it was used. I add these changes the original commit so they are all in one commit.

Levi-Armstrong:
I think everything is address except the formatting issues, which I will fix in a followup PR using clang format.

rct_common/include/rct_common/extrinsic_data_set.h
marip8: Let's delete this file

Levi-Armstrong:
@marip8 I created a PR add the linters which I will rebase and run after this is merged.

marip8:
Can you remove the empty extrinsic data header file? It still looks to be added in this PR. Otherwise I think it should be ready to merge

Levi-Armstrong:
@marip8 I removed that file.

rct_optimizations/include/rct_optimizations/covariance_analysis.h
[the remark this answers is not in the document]
Levi-Armstrong: Yea, not sure if there is a better way to handle this. When creating the problem we use the mask data and if all parameters in a group are masked it sets it to constant. Though I could possible remove this check and solely rely on the mask data. @marip8 What do you thing?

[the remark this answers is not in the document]
marip8: I'm inclined to say that we remove this check and let the user explicitly provide the mask. The covariance between a variable and a constant is 0, so worst case scenario is that you'll see a bunch of zeros if you forget to mask a constant block. 

Following the idea of autmatically removing constant parameters from the covariance: another thing that makes this tricky is that individual parameters can be held constant using a `SubsetParameterization`. We could also automatically remove those from the covariance matrix (without specifiying a mask) by calling `GetCovarianceMatrixInTangentSpace` instead of `GetCovarianceMatrix`. At this point however, the size/indices of your covariance matrix no longer correspond to the parameter block labels and indices. This would also be made worse if any other local parameterizations were used (e.g., the local parameterization that represents quaternions with 3 values instead of 4).

TLDR: let's get the full covariance matrix (like we're already doing), not automatically remove constant blocks, and rely explicitly on the masks to show the user the covariances they asked to see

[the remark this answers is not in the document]
Levi-Armstrong: I just added a commit to only use the mask info.

marip8 (approved):

Levi-Armstrong:
@marip8 Thanks for reviewing this!

PR #104, opened 2022-05-09
Improve finding the origin of a modified circle grid

marrts:
Compare corner relative sizes rather than absolute sizes

### DIFF
--- a/rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
+++ b/rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
@@ -60,11 +60,6 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
   std::vector<cv::KeyPoint> keypoints;
   cv::Point large_point;
 
-  std::size_t start_first_row = 0;
-  std::size_t end_first_row = cols - 1;
-  std::size_t start_last_row = rows * cols - cols;
-  std::size_t end_last_row = rows * cols - 1;
-
   for (double alpha = 1.0; alpha <= 3.0; alpha += 0.01)
   {
     bool found = false;
@@ -106,16 +101,30 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
   std::size_t temp_rows = flipped ? cols : rows;
   std::size_t temp_cols = flipped ? rows : cols;
 
+  std::size_t start_first_row = 0;
+  std::size_t end_first_row = cols - 1;
+  std::size_t start_last_row = rows * cols - cols;
+  std::size_t end_last_row = rows * cols - 1;
+
   // Determine which circle is the largest
   double start_first_row_size = -1.0;
   double start_last_row_size = -1.0;
   double end_first_row_size = -1.0;
   double end_last_row_size = -1.0;
 
+  // Determine size of corners relative to their adjacent circles
+  double start_first_row_avg_rel_size = 0;
+  double start_last_row_avg_rel_size = 0;
+  double end_first_row_avg_rel_size = 0;
+  double end_last_row_avg_rel_size = 0;
+
   cv::Point2d start_last_row_pt = centers[start_last_row];
   cv::Point2d end_last_row_pt = centers[end_last_row];
   cv::Point2d start_first_row_pt = centers[start_first_row];
   cv::Point2d end_first_row_pt = centers[end_first_row];
+
+  // Get the size of every circle
+  std::vector<double> center_sizes(centers.size(), 1.0);
 
   for (std::size_t i = 0; i < keypoints.size(); i++)
   {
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
@marip8 You have any issues with this change?

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
+  double end_first_row_avg_rel_size = 0;
+  double end_last_row_avg_rel_size = 0;
+
   cv::Point2d start_last_row_pt = centers[start_last_row];
   cv::Point2d end_last_row_pt = centers[end_last_row];
   cv::Point2d start_first_row_pt = centers[start_first_row];
   cv::Point2d end_first_row_pt = centers[end_first_row];
 
+  // Get the size of every circle
+  std::vector<double> center_sizes(centers.size(), 1.0);
```
marip8, line 127: It seems strange to seed this with an arbitrary number. Is that necessary for some reason?

```suggestion
  std::vector<double> center_sizes(centers.size());
```

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
   double start_first_row_size = -1.0;
   double start_last_row_size = -1.0;
   double end_first_row_size = -1.0;
   double end_last_row_size = -1.0;
 
+  // Determine size of corners relative to their adjacent circles
+  double start_first_row_avg_rel_size = 0;
+  double start_last_row_avg_rel_size = 0;
+  double end_first_row_avg_rel_size = 0;
+  double end_last_row_avg_rel_size = 0;
```
marip8, line 119:
```suggestion
  double start_first_row_avg_rel_size = 0.0;
  double start_last_row_avg_rel_size = 0.0;
  double end_first_row_avg_rel_size = 0.0;
  double end_last_row_avg_rel_size = 0.0;
```

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
+
   for (std::size_t i = 0; i < keypoints.size(); i++)
   {
     double x = keypoints[i].pt.x;
     double y = keypoints[i].pt.y;
     double ksize = keypoints[i].size;
 
+    for (std::size_t j = 0; j < centers.size(); j++)
+    {
+      auto center = centers[j];
```
marip8, line 137: Don't copy unnecessarily
```suggestion
      const auto& center = centers[j];
```

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
+      if (center.x == x && center.y == y)
```
marip8, line 138: Float equality comparison is a bad idea because it's unlikely that numbers will be the same out to the full precision. We should be checking equivalence instead (i.e. `abs(center.x - x) < std::numeric_limits<double>::epsilon()`). This seems to be a recurring elsewhere in this file as well, so we should probably fix it now. I would suggest making a helper function to do this equivalence comparison

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
+  // Determine size of corners relative to their adjacent circles
+  double start_first_row_avg_rel_size = 0;
+  double start_last_row_avg_rel_size = 0;
+  double end_first_row_avg_rel_size = 0;
+  double end_last_row_avg_rel_size = 0;
+
   cv::Point2d start_last_row_pt = centers[start_last_row];
   cv::Point2d end_last_row_pt = centers[end_last_row];
   cv::Point2d start_first_row_pt = centers[start_first_row];
   cv::Point2d end_first_row_pt = centers[end_first_row];
```
marip8, line 124:
```suggestion
  const cv::Point2d& start_last_row_pt = centers[start_last_row];
  const cv::Point2d& end_last_row_pt = centers[end_last_row];
  const cv::Point2d& start_first_row_pt = centers[start_first_row];
  const cv::Point2d& end_first_row_pt = centers[end_first_row];
```

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
     }
     if (x == end_first_row_pt.x && y == end_first_row_pt.y)
     {
       end_first_row_size = ksize;
+      if (i - 1 < center_sizes.size() && i + cols < center_sizes.size())
+      {
+        double end_first_row_rel_row_size = end_first_row_size / center_sizes[i-1];
+        double end_first_row_rel_col_size = end_first_row_size / center_sizes[i+cols];
+        end_first_row_avg_rel_size = (end_first_row_rel_row_size + end_first_row_rel_col_size)/2.0;
+      }
```
marip8, line 191: Please add some comments about what exactly you are comparing here to get the average relative size. I don't exactly follow which centers you are picking in each case for comparison

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
marip8, line 127: The name `center_sizes` is a bit misleading. `centers` is an argument that defines the locations of the features in the calibration target. These features have an actual size on the pattern. However, this particular variable actually represents the size of the observed keypoint features that correspond to the features on the calibration target. I would suggest renaming the argument `centers` to `target_features` and renaming to something like `obs_feature_sizes` to be more clear. It's a little outside the scope of this PR, but if we're making changes, we should make them clear

marip8 (changes requested):

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
marrts: I later divide by this number and didn't want to accidentally divide by zero

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -106,38 +101,94 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
...
     double y = keypoints[i].pt.y;
     double ksize = keypoints[i].size;
 
+    for (std::size_t j = 0; j < centers.size(); j++)
+    {
+      auto center = centers[j];
+      if (center.x == x && center.y == y)
+      {
+        center_sizes[j] = ksize;
+        break;
```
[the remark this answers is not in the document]
marrts: Wouldn't it break out of the for loop outside the if statement?

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
marrts: Renamed `centers` to `feature_coordinates` and `center_sizes` to `feature_sizes`, this mimics the keypoints definitions more closely because in  the `KeyPoint` struct they refer to the `pt` field as the "coordinates of the keypoints".

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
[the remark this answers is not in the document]
marip8: A little online searching seems to indicate that you're right. I initially thought break statements only work for the scope that they are in, but I think they are specifically for loops

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -3,6 +3,12 @@
 #include <opencv2/calib3d.hpp>
 #include <memory>
 
+template <typename T>
+bool isEqual(const T& a, const T& b)
+{
+    return abs(a - b) < std::numeric_limits<T>::epsilon();
+}
```
marip8, line 10: Being a little pedantic, but this function checks approximate equality. It might also be beneficial to allow a tolerance parameter in case the machine epsilon is too tight

```suggestion
template <typename T>
bool isApproxEqual(const T& a, const T& b, const T& tol = std::numeric_limits<T>::epsilon())
{
    return abs(a - b) < tol;
}
```

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
```diff
@@ -85,65 +86,90 @@ static std::vector<cv::Point2d> extractKeyPoints(const cv::Mat& image, const std
     }
   }
 
-  // OpenCV creates duplicate centers sometimes.  https://github.com/opencv/opencv/issues/4775
-  // Make sure every center has a matching keypoint.  It may be sufficient just to check centers for duplicates.
-  std::vector<cv::Point2d> centers_tmp = centers;
+  // OpenCV creates duplicate feature_coordinates sometimes.  https://github.com/opencv/opencv/issues/4775
+  // Make sure every center has a matching keypoint.  It may be sufficient just to check feature_coordinates for duplicates.
+  std::vector<cv::Point2d> feature_coordinates_tmp = feature_coordinates;
```
marip8, line 91: It seems like `feature_coordinates_tmp` gets created here, but `feature_coordinates` still gets used in the later parts of this function. We should probably change the function signature to take `feature_coordinates` by value so we can modify it in place and then remove `feature_coordinates_tmp`. That way we don't accidentally one of the variables some places and the other variable in other places by mistake

marip8 (changes requested):

rct_image_tools/src/rct_image_tools/modified_circle_grid_finder.cpp
marrts: The only purpose of `feature_coordinates_tmp` is that for loop of keypoints to check that every feature has a corresponding keypoint, that's why its made as a tmp and then it is empty a few lines later so it can't be used later

marip8 (approved):
Looks good to me. I'll merge once it passes CI

PR #110, opened 2023-11-13
Modified circle grid finder update

marip8:
Changes from @marrts 09f605f9ed06fa436d04c486129ad65e0b14abf1 and 5fbd140e349f7ef76b04f5902c663b06a6b733bf to improve the performance of the modified circle grid target finder


marip8:
@marrts aren't most of these `==` comparisons between doubles? If so, we should probably change them to the `isEqual` function instead

marrts:
> @marrts aren't most of these `==` comparisons between doubles? If so, we should probably change them to the `isEqual` function instead

They are double comparisons, but it is iterating through a vector and comparing to values that were previously extracted from that same vector, so it's maybe not the optimal approach for what it's trying to do, but the `==` comparison should always evaluate to `true` when it finds the right value

marip8:
Closing; turns out, the reference commits were added and modified in #104, which has the correct changes

PR #115, opened 2024-09-26
Update for Ceres 2.2 and Ubuntu Noble

schornakj:
Hello!

I'm using this project on Ubuntu 24.04, which requires at least Ceres 2.2. These are the changes I needed to add to get the non-ROS part of this repo building (I'm using ROS 2 Jazzy so I haven't looked at the ROS 1 packages in this repo yet to see if they need to change).

- Update uses of some Ceres functions that already were deprecated or soon will be:
  - `IsNaN` -> `isnan`
  - `GetParameterization` -> `GetManifold`
- The minimum C++ version needs to be updated to C++17.

This raises a repo version management question: how should we handle changes that break compatibility with older OS versions? In other projects I've used different branches for each target platform (`focal`, `noble`, etc.). In contrast I think [Tesseract](https://github.com/tesseract-robotics/tesseract) just drops support for older platforms as the tagged release advances.

### DIFF
--- a/rct_common/CMakeLists.txt
+++ b/rct_common/CMakeLists.txt
@@ -19,7 +19,7 @@ endif()
 find_gtest()
 
 add_library(${PROJECT_NAME} INTERFACE)
-target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_11)
+target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_17)
 target_include_directories(${PROJECT_NAME} INTERFACE "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
                                                      "$<INSTALL_INTERFACE:include>")
 target_link_libraries(${PROJECT_NAME} INTERFACE Eigen3::Eigen)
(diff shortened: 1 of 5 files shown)


marip8:
> This raises a repo version management question: how should we handle changes that break compatibility with older OS versions?

Dropping support is one option, but not preferred if we can help it somewhat easily. What I did in [industrial_calibration](https://github.com/ros-industrial/industrial_calibration/pull/197) was add a preprocessor check for the version of Ceres, and I was able to fairly easily support focal, jammy, and noble.

On that note, I would recommend checking out the updated version of [`industrial_calibration`](https://github.com/ros-industrial/industrial_calibration) instead of this repository. I've migrated all of the contents of RCT to `industrial_calibration` now, reorganized a bit for improved modularity, added GUIs for extrinsic hand-eye and intrinsic calibration, and added ROS-enabled data collection pipelines/utilties in separate `industrial_calibration_ros<1|2>` repos. I think it's in a better state than RCT at this point, and I would like to sunset development on RCT in favor of `industrial_calibration`.

